PR #1, opened 2017-09-18
Reformatting README for Terraform Registry

cgriggs01:
Hi there,

renamed the path of the source to the module registry instead of the GitHub repo

removed variables because all variable information is sourced from the variable and output files.

If you can merge this into master and then add another release/tag so that the module registry will update. After that we'll be able to verify this module.

Thanks
Chris


danisla:
LGTM, thanks.

PR #7, opened 2018-06-16
Adds availability_type and crash_safe_replication variables.

skinlayers:


skinlayers:
At a glance, `user_labels` also needs to be added, but I'm not sure what form terraform expects the "set of key/value" in.

PR #12, opened 2018-08-16
Updating readme

stekole:
The port for postgres was incorrect and set to the mysql port. This was updated along with removing extra postgres config from the end of the mysql line 
`./cloud_sql_proxy -instances=${PGSQL_CONN_NAME}**=tcp:5432**`


stekole:
I signed it!

PR #16, opened 2018-10-05
Add availabilty type, required for HA

skinlayers:


namusyaka:
Hi @skinlayers,
Sorry for the late response.
As you can see, #21 has been merged, so you can use the `avaiability_type` argument on the postgresql module.

skinlayers:
5 and 1/2 months to add a single option? I think we'll being using Ansible for these tasks in the future, as this is far too long to wait for such a small change.

skinlayers:
@namusyaka 7 months wait just to see an identical MR (https://github.com/GoogleCloudPlatform/terraform-google-sql-db/pull/34) merged? This is ridiculous!

namusyaka:
@skinlayers I'm very sorry for that. I apologize for the confusion that has caused many contributors, including you, by changing our policy.
At one point in time, it was decided not to maintain the past modules on an ongoing basis, but we will aggressively merge backwards compatibility and simple changes.

We take care that the same thing does not happen.
Thanks.

PR #19, opened 2018-10-30
added availability_type to sql_database_instance settings

jesusch:
would be good to have this upstream.
also updating this module on terraform registry would be a good thing


skinlayers:
Dupe of https://github.com/GoogleCloudPlatform/terraform-google-sql-db/pull/16/files

PR #21, opened 2018-11-16
support high availability, add inspec integration

namusyaka:
As we talked in the internal, I created a PR for implementing high availability settings and the inspec integration.
Before reviewing this implementation, please review the module API design at first, the actual usages are in the `test/fixtures` directory.


namusyaka:
@danisla In addition, I'd also like @morgante to review this. Please send him a review request if possible.

skinlayers:
This is the 3rd pending PR to add HA. Do we know why none have been merged so far?
https://github.com/GoogleCloudPlatform/terraform-google-sql-db/pull/16/files

namusyaka:
@beatcracker Glad to see your comment.

@skinlayers Your change is not same with this pull request completely, but I think it looks reasonable to me.

@danisla Can we merge #16 ? Of course, the `availability_type` argument will also be supported by this pull request however we can merge #16 without long-term reviewing.
 I can follow up the upstream after merging that.

modules/mysql/main.tf
```diff
@@ -0,0 +1,194 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+locals {
+  default_tier                            = "db-n1-standard-1"
```
morgante: Defaults should be encoded in variables instead of locals (if possible). Is there a reason these can't be?

modules/mysql/main.tf
```diff
@@ -0,0 +1,194 @@
...
+    }
+  }
+
+  lifecycle {
+    ignore_changes = ["disk_size"]
+  }
+}
+
+resource "google_sql_database_instance" "replicas" {
+  count                 = "${lookup(var.read_replica, "length", 0)}"
```
morgante: It looks like read_replica is a map, not an array? So I think this logic is faulty - instead we should just have a variable which specifies whether a read replica is desired or not (HA)?

modules/mysql/main.tf
```diff
@@ -0,0 +1,194 @@
...
+      update_track = "${lookup(var.master, "maintenance_window_update_track", local.default_maintenance_window_update_track)}"
+    }
+  }
+
+  lifecycle {
+    ignore_changes = ["disk_size"]
+  }
+}
+
+resource "google_sql_database_instance" "replicas" {
```
morgante: Let's separate the main, read replicas, and failovers into their own files for ease of review.

modules/mysql/main.tf
```diff
@@ -0,0 +1,194 @@
...
+  }
+
+  depends_on = ["google_sql_database_instance.master"]
+
+  lifecycle {
+    ignore_changes = ["disk_size"]
+  }
+}
+
+resource "google_sql_database" "databases" {
```
morgante: Since databases functionality isn't provided today, I think let's save this for a follow-on PR.

modules/mysql/outputs.tf
```diff
@@ -0,0 +1,128 @@
...
+  description = "The CA certificate information used to connect to the SQL instance via SSL"
+}
+
+output "master_instance_service_account_email_address" {
+  value       = "${google_sql_database_instance.master.service_account_email_address}"
+  description = "The service account email address assigned to the master instance"
+}
+
+// Replicas
+output "replicas_instance_first_ip_addresses" {
```
morgante: Why is this the first instance? The splat syntax will fetch all of them.

modules/mysql/outputs.tf
```diff
@@ -0,0 +1,128 @@
...
+  value       = ["${google_sql_database_instance.replicas.*.ip_address}"]
+  description = "The first IPv4 addresses of the addesses assigned for the replica instances"
```
morgante:
```suggestion
  description = "The first IPv4 addresses of the addresses assigned for the replica instances"
```

modules/mysql/variables.tf
```diff
@@ -0,0 +1,212 @@
...
+
+// required
+variable "region" {
+  description = "The region of the Cloud SQL resources"
+}
+
+variable "master" {
+  default = {}
+
+  description = <<EOF
```
morgante: Let's split all of these out into `master_*` variables. It makes configuration and defaults easier to manage.

modules/mysql/variables.tf
```diff
@@ -0,0 +1,212 @@
...
+
+variable "master_labels" {
+  default     = {}
+  description = "The key/value labels for the master instances."
+}
+
+variable "read_replica" {
+  default = {}
+
+  description = <<EOF
```
morgante: Let's split this into `read_replica_*` variables. Ex `read_replica_tier`.

modules/mysql/variables.tf
```diff
@@ -0,0 +1,212 @@
...
+variable "read_replica_labels" {
+  default     = {}
+  description = "The key/value labels for the read replica instances."
+}
+
+variable "failover_replica" {
+  default = {}
+
+  description = <<EOF
+The failover replica settings of the Cloud SQL resources
```
morgante: Ditto, split this into `failover_replica_*`

examples/mysql-and-postgres/README.md
```diff
@@ -1,70 +0,0 @@
-# Cloud SQL Database Example
```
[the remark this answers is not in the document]
namusyaka: OK reverted this: 58d4f3a147e9e426c82e0bf536a0fa246ff0a1dd
And updated `main.tf` to make it act with new module: 39bca8011fbe294e6a89a39fdfaf5910ae3054e7

modules/mysql/main.tf
```diff
@@ -0,0 +1,192 @@
...
+    disabled = "${list(local.primary_zone)}"
+  }
+
+  zones_enabled = "${length(local.read_replica_zones) > 0}"
+  mod_by        = "${local.zones_enabled ? length(local.read_replica_zones) : 1}"
+
+  zones = "${local.zone_mapping["${local.zones_enabled ? "enabled" : "disabled"}"]}"
+}
+
+resource "google_sql_database_instance" "master" {
```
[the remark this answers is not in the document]
namusyaka: Ack. Fixed both mysql and postgres: 0982e59ce16aa38c46893e0d0a7b786c774ee97b

modules/mysql/main.tf
```diff
@@ -0,0 +1,192 @@
...
+  project          = "${var.project_id}"
+  name             = "${var.name}"
+  database_version = "${var.database_version}"
+  region           = "${var.region}"
+
+  settings {
+    tier                        = "${lookup(var.master, "tier", local.default_tier)}"
```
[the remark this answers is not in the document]
namusyaka: To improve readability, but now we have some `master_*` arguments and current situation is odd, so I agreed about split fields into each arguments. But the `master_` prefix is not required because the `master` instance is now provided as `default`.
See: d86bddbf9ff3c732d16897e2c596ea5572f47910

modules/mysql/main.tf
namusyaka: Sorry I can't remember that because it has passed a little while since I sent a PR.
Probably your review seems good. Moved: d86bddbf9ff3c732d16897e2c596ea5572f47910 and 5f102ec06513661a0030203a61e72ff534f6a0c4

modules/mysql/main.tf
namusyaka: Nope, this works fine. As you said, `read_replica` is map. But if read replica is not required, length will be 0 by default because the developer won't specify the `read_replica` argument.
`read_replica`'s arguments will be used for all read replicas as settings, only `zones` argument will be used for rotation.

Receiving your comment, I just added a few arguments for improving readability, they're `failover_replica`, `read_replica_size`.
The `failover_replica` is expected to set boolean. The `read_replica_size` is same with the `length` argument inside the `read_replicas` block.

Please review again.

modules/mysql/main.tf
namusyaka: Well, it should be supported by [Databases: insert](https://cloud.google.com/sql/docs/mysql/admin-api/v1beta4/databases/insert?hl=en).
In fact, multiple databases can be created by using this resource.

modules/mysql/outputs.tf
namusyaka: Fixed in: 0982e59ce16aa38c46893e0d0a7b786c774ee97b

namusyaka:
I have reflected reviews, thanks @morgante!
I'm wondering how far backward compatibility should be concerned. To be able to upgrade at least smoothly, argument names are unified as much as possible, but maintaining full compatibility is difficult..

namusyaka:
I have reflected reviews and replied to a few comments, thanks @morgante!
I am wondering how far backward compatibility should be concerned. To be able to upgrade at least smoothly, argument names are unified as much as possible, but maintaining full compatibility is difficult.

In addition, all tests are passed, of course.

morgante:
@namusyaka Regarding backward compatibility, the biggest thing is to avoid state migrations (usually caused by renaming resources). If arguments slightly change, that's less of an issue.

namusyaka:
@morgante I did took care to reduce the resource level changes.

skinlayers:
@namusyaka And this is why I was hoping my atomic change would get merged. This PR includes a lot more than just HA, and still seems to be under development.

namusyaka:
@skinlayers This is a major refurbishment for the next version of this terraform module. This is not just adding availability_type, but aims to upgrade to a module that can realize high availability for production environment.
I said that your pr is reasonable because it already existed.

This module is basically in the review stage, especially no additional development is assumed.

skinlayers:
Can we get an update? There's been no visible movement on this in the last 18 days. If this major update isn't being currently worked on, then please merge my PR so we don't have to keep waiting and wondering when this will get done.

namusyaka:
@danisla @morgante ping

modules/mysql/main.tf
namusyaka: @morgante Is this fine?

modules/mysql/main.tf
namusyaka: That makes sense. To avoid big migration, I think we should also switch `google_sql_database` and `google_sql_user` back to `default`.

namusyaka: Fixed: https://github.com/GoogleCloudPlatform/terraform-google-sql-db/pull/21/commits/de16a90bcde43968ab888850a72f4962fa40232e

namusyaka:
Thank you both. I'll send a few PRs for following up on this.

PR #26, opened 2019-03-05
Updates the hcl linting

geojaz:
Lints all the terraform in the repo and fixes if needed. Perhaps I should change check_terraform to validate_terraform?

The second commit is after everything is applied.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -31,16 +31,5 @@ be able to accept your pull requests.
 
 ## Style
 
-Format your HCL code with [`hclfmt`](https://github.com/fatih/hclfmt).
-
-Make sure there are no differences between the `hclfmt` output and your .tf files:
-
-```
-diff -B <( cat *.tf ) <( hclfmt *.tf )
-```
-
-Write changes to the files:
-
-```
-hclfmt -w *.tf
-```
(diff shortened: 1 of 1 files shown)


test/make.sh
```diff
@@ -49,8 +49,20 @@ function check_terraform() {
...
   find . -name "*.tf" -exec bash -c 'terraform validate --check-variables=false $(dirname "{}")' \;
-  echo "Running terraform fmt"
-  terraform fmt -check=true -write=false
+}
+
+# This function lints terraform files (ending with *.tf) with hclfmt
+function lint_terraform() {
+  echo "Checking and fixing terraform formatting"
+  #shellcheck disable=SC2046
+  diff -B <(cat $(find . -name "*.tf")) <(hclfmt $(find . -name "*.tf"))
```
namusyaka: I'm not familiar with `hclfmt` but I confirmed it's no longer maintained.
https://github.com/fatih/hclfmt

If we would like to adopt `hclfmt` into this repository, we need to mention it in [our README](https://github.com/GoogleCloudPlatform/terraform-google-sql-db/blob/master/README.md).

morgante: Specifically, what's the reasoning for switching to `hclfmt` over `terraform fmt`?

geojaz:
It looks like there was some out of date information to use hcl_fmt in the contributing.md. Updated it to explain how to get your code linted.

PR #28, opened 2019-03-18
Exposes timeout parameters

geojaz:


geojaz:
@danisla @namusyaka could you please take a look at this and help as get these parameters exposed? We're using this module for some helmsman demos and we need this functionality. Thanks for your help!

modules/mysql/main.tf
```diff
@@ -59,6 +59,12 @@ resource "google_sql_database_instance" "default" {
   lifecycle {
     ignore_changes = ["disk_size"]
   }
+
+    timeouts {
+    create = "${var.create_timeout}"
+    update = "${var.update_timeout}"
+    delete = "${var.delete_timeout}"
+  }
 }
```
namusyaka: Have you checked `modules/{mysql,postgresql}/{read_replica,failover_replica}.tf`?
If you would like to manage the timeout of heavy processing, you should specify the `timeouts` configuration block in other `google_sql_instance` resources as well.

modules/postgresql/README.md
```diff
@@ -6,69 +6,72 @@
...
+| instance\_name | The instance name for the master instance |
+| instance\_self\_link | The URI of the master instance |
+| instance\_server\_ca\_cert | The CA certificate information used to connect to the SQL instance via SSL |
+| instance\_service\_account\_email\_address | The service account email address assigned to the master instance |
+| read\_replica\_instance\_names | The instance names for the read replica instances |
+| replicas\_instance\_connection\_names | The connection names of the replica instances to be used in connection strings |
+| replicas\_instance\_ip\_addresses | The IPv4 addresses assigned for the replica instances |
+| replicas\_instance\_self\_links | The URIs of the replica instances |
+| replicas\_instance\_server\_ca\_certs | The CA certificates information used to connect to the replica instances via SSL |
+| replicas\_instance\_service\_account\_email\_addresses | The service account email addresses assigned to the replica instances |
```
namusyaka: I'm aware of this changes because I reviewed #26 however you should drop the changes generated by the lint at this time.
We should discuss the subject of the lint on #26

geojaz:
@namusyaka thanks for the feedback. I think we should be fixed up on this one.

namusyaka:
Changes in `.tf` seems good, but I'm wondering why the underscores are escaped. Could you elaborate?

geojaz:
@namusyaka i think my last update should have gotten it.

namusyaka:
As we mentioned in README, make generate_docs should generate variables table automatically. Did you use it?
Current change looks like still changing docs manually.

geojaz:
I pushed another iteration on this where I went back to the files as per master branch and generated from them... it looks like the behavior of terraform-docs may have changed (https://github.com/segmentio/terraform-docs/issues/48) which is probably why the READMEs looked manually generated/funky.

I think this is right? If there's a different set of tools that I should be using, happy to get those installed and update docs if needed. Sorry this keeps going round and round :(

namusyaka:
OK great, I'm going to verify the current behavior of `terraform-docs` then I will merge this soon.
Thanks for your contribution.

geojaz:
awesome thanks for helping see this through!

PR #29, opened 2019-04-01
Added availability_type

ingvarch:


namusyaka:
@ingvarch Thanks for the pull request. Btw, can you use separated modules instead of this?

ingvarch:
@namusyaka I think so, but which one?

namusyaka:
The module has been updated drastically in #21.
In the PR, the main module you updated is separated into two modules, [mysql](https://github.com/GoogleCloudPlatform/terraform-google-sql-db/tree/master/modules/mysql) and [postgresql](https://github.com/GoogleCloudPlatform/terraform-google-sql-db/tree/master/modules/postgresql).
The `availability_type` is available only on the postgresql module because [it's available only for PostgreSQL](https://www.terraform.io/docs/providers/google/r/sql_database_instance.html).

I don't think we need to add the variable in our old terraform module.

ingvarch:
Ah, I see. Ok :) Thank a lot. Close PR.

PR #35, opened 2019-05-17
Added a module for MySQL for configurations that are safer than the default

mmontan:
The added module creates an instance with two main properties:
- all connections are authorized via Cloud IAM
- has a private IP for communication within the project.

We maintain the option of binding the instance to a public IP to simplify debugging and inter-process communication when using only a private ip is not possible. The public IP doesn't have any authorized networks, so all connections need to be mediated by the Cloud SQL proxy.

==
A couple of notes:
- I changed the way in which project name is passed in the example implementation. The use of the default project name didn't work for my gcloud, even if I had a default project set. Maybe some change in behavior?

- I removed a complicated setup for ip_connections that was giving errors in master. I don't understand why it was done that way, but the current solution can handle both passing ip_configuration passed as a list, or passed as simple object.  PTAL and make sure I am not breaking something because of it.

- The InSpec for the safer-mysql module doesn't check for the properties such as ip_connections. Any suggestion on how to generate the expected configuration?


mmontan:
@morgante @yukinying PTAL ( I can't add reviewers)

namusyaka:
Will review next week.
Thanks for the PR!

modules/mysql/main.tf
```diff
@@ -35,7 +29,7 @@ resource "google_sql_database_instance" "default" {
     activation_policy           = "${var.activation_policy}"
     authorized_gae_applications = ["${var.authorized_gae_applications}"]
     backup_configuration        = ["${var.backup_configuration}"]
-    ip_configuration            = "${local.ip_configurations["${local.ip_configuration_enabled ? "enabled" : "disabled"}"]}"
+    ip_configuration            = ["${var.ip_configuration}"]
```
namusyaka: By this change, `make test_integration` fails, because this still recognizes empty `ip_configuration` even if `ip_configuration` is unspecified.

Current behavior:
```
         settings.#:                                   "" => "1"
         settings.0.activation_policy:                 "" => "ALWAYS"
         settings.0.availability_type:                 "" => "<computed>"
         settings.0.backup_configuration.#:            "" => "1"
         settings.0.backup_configuration.0.start_time: "" => "<computed>"
         settings.0.crash_safe_replication:            "" => "<computed>"
         settings.0.disk_autoresize:                   "" => "true"
         settings.0.disk_size:                         "" => "10"
         settings.0.disk_type:                         "" => "PD_SSD"
         settings.0.location_preference.#:             "" => "1"
         settings.0.location_preference.0.zone:        "" => "us-central1-c"
         settings.0.maintenance_window.#:              "" => "1"
         settings.0.maintenance_window.0.day:          "" => "1"
         settings.0.maintenance_window.0.hour:         "" => "23"
         settings.0.maintenance_window.0.update_track: "" => "canary"
         settings.0.pricing_plan:                      "" => "PER_USE"
         settings.0.replication_type:                  "" => "SYNCHRONOUS"
         settings.0.tier:                              "" => "db-n1-standard-1"
         settings.0.version:                           "" => "<computed>"
```

Your changes:
```
         settings.#:                                   "" => "1"
         settings.0.activation_policy:                 "" => "ALWAYS"
         settings.0.availability_type:                 "" => "<computed>"
         settings.0.backup_configuration.#:            "" => "1"
         settings.0.backup_configuration.0.start_time: "" => "<computed>"
         settings.0.crash_safe_replication:            "" => "<computed>"
         settings.0.disk_autoresize:                   "" => "true"
         settings.0.disk_size:                         "" => "10"
         settings.0.disk_type:                         "" => "PD_SSD"
         settings.0.ip_configuration.#:                "" => "1" # <- PTAL
         settings.0.ip_configuration.0.ipv4_enabled:   "" => "<computed>"
         settings.0.location_preference.#:             "" => "1"
         settings.0.location_preference.0.zone:        "" => "us-central1-c"
         settings.0.maintenance_window.#:              "" => "1"
         settings.0.maintenance_window.0.day:          "" => "1"
         settings.0.maintenance_window.0.hour:         "" => "23"
         settings.0.maintenance_window.0.update_track: "" => "canary"
         settings.0.pricing_plan:                      "" => "PER_USE"
         settings.0.replication_type:                  "" => "SYNCHRONOUS"
         settings.0.tier:                              "" => "db-n1-standard-1"
         settings.0.version:                           "" => "<computed>"
```

I would like to revert this changes.

modules/postgresql/main.tf
```diff
@@ -35,7 +26,7 @@ resource "google_sql_database_instance" "default" {
     availability_type           = "${var.availability_type}"
     authorized_gae_applications = ["${var.authorized_gae_applications}"]
     backup_configuration        = ["${var.backup_configuration}"]
-    ip_configuration            = "${local.ip_configurations["${local.ip_configuration_enabled ? "enabled" : "disabled"}"]}"
+    ip_configuration            = ["${var.ip_configuration}"]
```
namusyaka: Same as above.

test/fixtures/safer-mysql-simple/main.tf
```diff
@@ -0,0 +1,86 @@
...
+  database_version = "MYSQL_5_7"
+  region           = "us-central1"
+  zone             = "c"
+  tier             = "db-n1-standard-1"
+
+  # By default, all users will be permitted to connect only via the
+  # Cloud SQL proxy.
+  additional_users = [{
+    name = "app"
+  },
```
namusyaka: indent

mmontan: Fixed

modules/postgresql/main.tf
mmontan: Restored the behavior, but I still had to change this line.

Without the '[...]', the example instantiation fails. This solution passes the integration tests.

modules/mysql/main.tf
mmontan: Done. The new change restores the old behavior.

```
  connection_name:                              "" => "<computed>"
         database_version:                             "" => "POSTGRES_9_6"
         first_ip_address:                             "" => "<computed>"
         ip_address.#:                                 "" => "<computed>"
         master_instance_name:                         "" => "<computed>"
         name:                                         "" => "tf-pg-simple-x"
         private_ip_address:                           "" => "<computed>"
         project:                                      "" => "mmontan-cloud"
         public_ip_address:                            "" => "<computed>"
         region:                                       "" => "us-central1"
         replica_configuration.#:                      "" => "<computed>"
         self_link:                                    "" => "<computed>"
         server_ca_cert.#:                             "" => "<computed>"
         service_account_email_address:                "" => "<computed>"
         settings.#:                                   "" => "1"
         settings.0.activation_policy:                 "" => "ALWAYS"
         settings.0.availability_type:                 "" => "ZONAL"
         settings.0.backup_configuration.#:            "" => "1"
         settings.0.backup_configuration.0.start_time: "" => "<computed>"
         settings.0.crash_safe_replication:            "" => "<computed>"
         settings.0.disk_autoresize:                   "" => "true"
         settings.0.disk_size:                         "" => "10"
         settings.0.disk_type:                         "" => "PD_SSD"
         settings.0.location_preference.#:             "" => "1"
         settings.0.location_preference.0.zone:        "" => "us-central1-c"
         settings.0.maintenance_window.#:              "" => "1"
         settings.0.maintenance_window.0.day:          "" => "1"
         settings.0.maintenance_window.0.hour:         "" => "23"
         settings.0.maintenance_window.0.update_track: "" => "canary"
         settings.0.pricing_plan:                      "" => "PER_USE"
         settings.0.replication_type:                  "" => "SYNCHRONOUS"
         settings.0.tier:                              "" => "db-f1-micro"
         settings.0.version:                           "" => "<computed>"
```

mmontan:
I addressed the comments. 

Another issue that I'd like to look at is the change in the example where I had to add an explicit request for a project name. I am getting the same problem in master. However, I can remove it from this commit to simplify and send it as a separate pull request.

test/fixtures/safer-mysql-simple/main.tf
```diff
@@ -0,0 +1,86 @@
...
+    {
+      name = "readonly"
+    },
+  ]
+
+  assign_public_ip = "true"
+  vpc_network      = "${google_compute_network.default.self_link}"
+
+  # Optional, but used to enforce ordering in the creation of resources.
+  vpc_peering = "${google_service_networking_connection.private_vpc_connection.self_link}"
```
namusyaka: I think we should add `create_timeout` to avoid failing test by timeout.
Or we should update default timeout values in `variables.tf`

README.md
```diff
@@ -72,6 +72,9 @@ And then, you should pass the service account credentials for running inspec by
 
 - `GOOGLE_APPLICATION_CREDENTIALS`
 
+In addition to `roles/cloudsql.admin`, the service account used for testing needs
+`roles/compute.networkAdmin` to instantiate the VPC required for testing.
```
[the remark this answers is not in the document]
mmontan: Updated the README.md

test/fixtures/safer-mysql-simple/main.tf
mmontan: I updated the default value to 15 minutes.

namusyaka:
I let @morgante review this changes for double-checking.

examples/mysql-and-postgres/main.tf
```diff
@@ -87,29 +97,84 @@ module "postgresql-db" {
...
+  project       = "${var.project_id}"
+  name          = "private-ip-address"
+  purpose       = "VPC_PEERING"
+  address_type  = "INTERNAL"
+  prefix_length = 16
+  network       = "${google_compute_network.default.self_link}"
+}
+
+# Creates the peering with the producer network.
+resource "google_service_networking_connection" "private_vpc_connection" {
```
morgante: Should this eventually be included in the module itself?

modules/safer_mysql/README.md
```diff
@@ -0,0 +1,250 @@
+# terraform-google-sql for a safer MySQL
+
+The module sets up a MySQL installation that relies on Cloud IAM for
+authenticating its users. The use of Cloud IAM centralises identty
```
morgante:
```suggestion
authenticating its users. The use of Cloud IAM centralizes identity
```

modules/safer_mysql/README.md
```diff
@@ -0,0 +1,250 @@
...
+hence, be authorized via Cloud IAM.
+
+The most secure setup doesn't assign a public IP to the instance
+(assign_public_ip = false), and permits connections only from within the
+given VPC. In such a setup, sharing the instance between projects requires
+setting up Shared VPCs, and direct mysql access from engineering workstations
+is not possible: any debugging needs to be performed via bastion hosts or
+custom tooling running on VMs connected to the VPC.
+
+If assign_public_ip = true, the instance will be assigned a public IP. However,
```
morgante: What is the use case for `assign_public_ip` on the safe module?

modules/safer_mysql/README.md
```diff
@@ -0,0 +1,250 @@
...
+    will have read/write permissions on data, but no ability of performing
+    structural modifications to the DB.
+
+    *   Multiple roles for different apps can be defined if more protection
+        against accidental errors is desired.
+
+-   **readonly**: permissions required for reading data in the database for
+    debugging or customer support.
+
+*   Once users are created, you need to assign them
```
morgante: (Later solution) Should we incorporate this SQL automatically?

modules/safer_mysql/README.md
```diff
@@ -0,0 +1,250 @@
...
+wget https://dl.google.com/cloudsql/cloud_sql_proxy.linux.amd64 -O cloud_sql_proxy
+chmod +x cloud_sql_proxy
+```
+
+Connections from engineers' workstations is directly possible only if the instance
+has been assiged a public ip (assign_public_ip = "true"). If only Private IPs are
+used, then access needs to be mediated by bastion hosts connected to the same VPC,
+or through custom UI or other tools.
+
+If public IP is avaible, engineers can use the following process to connect:
```
morgante:
```suggestion
If public IP is available, engineers can use the following process to connect:
```

modules/safer_mysql/variables.tf
```diff
@@ -0,0 +1,338 @@
...
+
+// required
+variable "database_version" {
+  description = "The database version to use"
+}
+
+// required
+variable "region" {
+  description = "The region of the Cloud SQL resources"
+  default     = "us-central1"
```
morgante: Let's not have a default region.

modules/safer_mysql/variables.tf
```diff
@@ -0,0 +1,338 @@
...
+}
+
+// required
+variable "vpc_network" {
+  description = "Existing VPC network to which instances are connected. The networks needs to be configured with https://cloud.google.com/vpc/docs/configure-private-services-access."
+}
+
+variable "vpc_peering" {
```
morgante: Could this be incorporated into the module itself, or possibly as a submodule?

modules/safer_mysql/variables.tf
```diff
@@ -0,0 +1,338 @@
...
+}
+
+variable "maintenance_window_hour" {
+  description = "The hour of day (0-23) maintenance window for the master instance maintenance."
+  default     = 23
+}
+
+variable "maintenance_window_update_track" {
+  description = "The update track of maintenance window for the master instance maintenance. Can be either `canary` or `stable`."
+  default     = "canary"
```
morgante: Would `stable` not be preferable for a secure/production-ready workload?

modules/safer_mysql/variables.tf
```diff
@@ -0,0 +1,338 @@
...
+  description = <<EOF
+The backup configuration block of the Cloud SQL resources
+This argument will be passed through the master instance directrly.
+
+See [more details](https://www.terraform.io/docs/providers/google/r/sql_database_instance.html).
+EOF
+}
+
+variable "assign_public_ip" {
+  description = "Set tp true if the master instance should also have a public IP."
```
morgante:
```suggestion
  description = "Set to true if the master instance should also have a public IP (less secure)."
```

modules/safer_mysql/main.tf
```diff
@@ -0,0 +1,118 @@
...
+  user_labels                     = "${var.user_labels}"
+  backup_configuration            = "${var.backup_configuration}"
+
+  ip_configuration = [
+    {
+      ipv4_enabled = "${var.assign_public_ip}"
+
+      # We never set authorized networks, we need all connections via the
+      # public IP to be mediated by Cloud SQL.
+      authorized_networks = []
```
OGProgrammer: Ehhh, I use `authorized_networks` in my scripts via the `ip_configuration` block. Why hard code this to empty?

OGProgrammer: nvm I see this is just for a new option.

mmontan: Correct, this is just a new opinionated setup with some "safer" defaults, it won't affect anything for the current users.

modules/safer_mysql/README.md
mmontan: The rationale is that there are certain networking setups that are hard to use with only private ips, and the mandated use of Cloud SQL Proxy for connections mitigates sufficient risks.

In particular, there are limits in the way the Cloud SQL peering works, as it doesn't allow transitive connections between peered networks. So if we have:

 P1 <--> Private Cloud SQL

and now we want to use the same instance from P2, one natural thing would be to peer its VPC with P1. However,  P2 won't be able to see Cloud SQL in this setup.

P2 <--> P1 <-X only P1 X-> Cloud SQL. 

The solution for this are shared VPCs, but adopting them often requires significant changes in the architecture, and there are opinions stating that Shared VPCs might end up eroding the security boundary defined by a project, and sharing too many resources by default.

Because we only expose a public IP with NO authorized networks, we ensure that Mysql is not directly exposed, but its connections are mediated by the Cloud SQL Proxy and authorized by Cloud IAM, making Cloud SQL to be similar to other Cloud APIs endpoints that are exposed to the public.

@yukinying did I miss anything?

modules/safer_mysql/variables.tf
mmontan: Removed.

modules/safer_mysql/variables.tf
mmontan: Changed to stable. What do you think of the default maintenance window? Should we force users to pick one instead?

modules/safer_mysql/variables.tf
mmontan: I defined it as a different submodule, and updated the example and tests to use it.

it can't be incorporated in the module itself, as it is created only once for all Cloud SQL instances associated to the same VPC.

examples/mysql-and-postgres/main.tf
mmontan: Defined as a submodule, see below

modules/safer_mysql/README.md
mmontan: Good idea, I'll look into it for a followup pull request.

modules/safer_mysql/README.md
morgante: Got it, thanks for the thorough explanation.

modules/safer_mysql/variables.tf
morgante: Default is fine.

modules/safer_mysql/README.md
mmontan: (resolving for now, I can work on that improvement independently from this PR)

examples/mysql-and-postgres/main.tf
namusyaka: @mmontan What is your motivation for creating a new submodule?

modules/safer_mysql/variables.tf
namusyaka: ah sorry, I overlooked this comment

examples/mysql-and-postgres/main.tf
mmontan: The peering and assignment of IPs can only be done once for each VPC,  and it will be used by all Cloud SQL instances with private ips in that VPC.

Because multiple Cloud SQL private instances can be added to the same VPC, it seems that it would be cleaner to have it as an independent module that it is created before the instances.
 
Do you have other ideas on how this can be done?

test/integration/safer-mysql-simple/controls/mysql.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  its(:backend_type)     { should eq 'SECOND_GEN' }
+  its(:database_version) { should eq 'MYSQL_5_7' }
+  its(:state)            { should eq 'RUNNABLE' }
+  its(:region)           { should eq 'us-central1' }
+  its(:gce_zone)         { should eq 'us-central1-c' }
+
+  it { expect(settings).to include(expected_settings) }
+  it { expect(location_preference).to include(kind: "sql#locationPreference", zone: "us-central1-c") }
+  it { expect(maintenance_window).to include(kind: "sql#maintenanceWindow", day: 1, hour: 23, update_track: "canary") }
```
namusyaka: Could you replace `canary` with `stable`?

mmontan: Done.

PR #37, opened 2019-06-07
Fix terraform list issue [WIP]

0tza:
Version used : 
Terraform v0.12.1
+ provider.google v2.8.0
+ provider.random v2.1.2

When hitting `terraform plan`, terraform complains about the shake of the data. The variables on `variable.tf` are a list that gets inserted into another list on `main.tf`.

**Error Message**

```
Error: Unsupported argument

  on db.tf line 131, in resource "google_sql_database_instance" "default":
 131:     location_preference         = "${var.location_preference}"

An argument named "location_preference" is not expected here. Did you mean to
define a block of type "location_preference"?
```

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -24,21 +24,21 @@ resource "google_sql_database_instance" "default" {
   settings {
     tier                        = "${var.tier}"
     activation_policy           = "${var.activation_policy}"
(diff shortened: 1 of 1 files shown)


morgante:
Terraform v0.12 isn't yet supported by this module, we will have to update it for 0.12 support in a batch.

PR #38, opened 2019-06-11
fix: settings.0.ip_configuration: expected a list

thefirstofthe300:


modules/mysql/main.tf
```diff
@@ -19,8 +19,8 @@ locals {
   ip_configuration_enabled = "${length(keys(var.ip_configuration)) > 0 ? true : false}"
 
   ip_configurations = {
-    enabled  = "${list(var.ip_configuration)}"
-    disabled = "${list()}"
+    enabled  = "${var.ip_configuration}"
+    disabled = "${map()}"
```
namusyaka: Assuming this change is correct, we also need to update postgre's one: https://github.com/GoogleCloudPlatform/terraform-google-sql-db/blob/02a3a5d1514cd082ab1afc3af6f19192cadeb76c/modules/postgresql/main.tf#L22

modules/mysql/main.tf
```diff
@@ -35,7 +35,7 @@ resource "google_sql_database_instance" "default" {
     activation_policy           = "${var.activation_policy}"
     authorized_gae_applications = ["${var.authorized_gae_applications}"]
     backup_configuration        = ["${var.backup_configuration}"]
-    ip_configuration            = "${local.ip_configurations["${local.ip_configuration_enabled ? "enabled" : "disabled"}"]}"
+    ip_configuration            = ["${local.ip_configurations["${local.ip_configuration_enabled ? "enabled" : "disabled"}"]}"]
```
namusyaka: We also need to update `failover_replica.tf` and `read_replica.tf`: https://github.com/GoogleCloudPlatform/terraform-google-sql-db/blob/02a3a5d1514cd082ab1afc3af6f19192cadeb76c/modules/mysql/read_replica.tf#L44 and https://github.com/GoogleCloudPlatform/terraform-google-sql-db/blob/master/modules/mysql/read_replica.tf#L44

namusyaka:
In my local, `test/fixtures/mysql-simple` fails after reflecting this changes.

modules/mysql/main.tf
thefirstofthe300: That change is correct.

PSQL has been updated as well.

modules/mysql/main.tf
thefirstofthe300: Updated.

thefirstofthe300:
@namusyaka All test modules have been updated and all tests are passing.

Can I get another review?

modules/postgresql/variables.tf
```diff
@@ -115,7 +115,10 @@ variable "authorized_gae_applications" {
 
 variable "ip_configuration" {
   description = "The ip configuration for the master instances."
-  default     = {}
+
+  default = {
+    ipv4_enabled = "true"
+  }
```
namusyaka: I think this fixes a broken test potentially, but is it on purpose?
In my opinion, this looks reasonable because our official document says:

```
The settings for IP Management. This allows to enable or disable the instance IP and manage which external networks can connect to the instance. The IPv4 address cannot be disabled for Second Generation instances.
```
https://cloud.google.com/sql/docs/mysql/admin-api/v1beta4/instances/insert#request-body

What do you think?

test/fixtures/postgresql-ha/main.tf
```diff
@@ -114,6 +114,10 @@ module "pg" {
   user_password = "foobar"
 
   additional_users = [
+    {
+      name     = "tftest1"
+      password = "raboof"
+    },
```
namusyaka: What's the motivation to add this ?

modules/postgresql/variables.tf
thefirstofthe300: This actually fixes the module being broken entirely:
```
ipv4_enabled - (Optional) Whether this Cloud SQL instance should be assigned a public IPV4 address. Either ipv4_enabled must be enabled or a private_network must be configured.
```
Passing in an empty map leads to an error due to Terraform expecting one of these values to be set.

test/fixtures/postgresql-ha/main.tf
thefirstofthe300: To fix the test logic. The test to verify the user count looks for four users with that start with `tftest`: https://github.com/GoogleCloudPlatform/terraform-google-sql-db/blob/627ae502386e263c1f48073a2b7ba9db408ee5fd/test/integration/postgresql-ha/controls/pg.rb#L99-L103

As it stands, the test only added three users and the user names are tftest, tftest2, and tftest3. I figured the omission of tftest1 was an oversight.

modules/postgresql/variables.tf
namusyaka: Yep I know, but previous implementation doesn't raise an error.
I think current implementation looks fine, thanks.

test/fixtures/postgresql-ha/main.tf
namusyaka: Nope it has already been fixed by https://github.com/GoogleCloudPlatform/terraform-google-sql-db/commit/aec7fdb6f3c0de16c64141b4979537f916573811

You should have rebased the master branch.
I'll revert this.

PR #43, opened 2019-07-03
Postgres support for private_service_access

dpetersen:
I am trying to use the `postgresql` submodule with the `private_service_access` submodule that was created as part of #35 for MySQL (really appreciate that work!). There's nothing in `private_service_access` that's MySQL specific, but the `peering_completed` variable workaround to enforce dependency order was only added to `safer_mysql`.

I also added a [`private_address` output](https://www.terraform.io/docs/providers/google/r/sql_database_instance.html#private_ip_address) to `postgresql`, trying to mirror the preexisting `instance_address` output. The README for the project says this works with a 1.12.x version of the google provider, but `private_ip_address` wasn't introduced until version [version 2.1.0](https://github.com/terraform-providers/terraform-provider-google/blob/master/CHANGELOG.md#210-february-26-2019). Does the README cover submodules, too? What's your policy on requiring a newer version of the provider?

I tried to stick as closely as possible to naming and descriptions to what was already in the codebase. Let me know if you'd like anything changed. With these changes I was able to use private networking on our postgres instance, which is awesome. Thanks!

### DIFF
--- a/modules/postgresql/main.tf
+++ b/modules/postgresql/main.tf
@@ -21,6 +21,13 @@ locals {
     enabled  = "${var.ip_configuration}"
     disabled = "${map()}"
   }
+
+  peering_completed_enabled = "${var.peering_completed != "" ? true : false}"
+
+  user_labels_including_tf_dependency = {
+    enabled  = "${merge(map("tf_dependency", var.peering_completed), var.user_labels)}"
+    disabled = "${var.user_labels}"
+  }
 }
 
 resource "google_sql_database_instance" "default" {
@@ -46,9 +53,7 @@ resource "google_sql_database_instance" "default" {
     // Define a label to force a dependency to the creation of the network peering.
     // Substitute this with a module dependency once the module is migrated to
     // Terraform 0.12
-    user_labels = "${merge(
-      map("tf_dependency", var.peering_completed),
-      var.user_labels)}"
+    user_labels = "${local.user_labels_including_tf_dependency["${local.peering_completed_enabled ? "enabled" : "disabled"}"]}"
 
     location_preference {
       zone = "${var.region}-${var.zone}"
--- a/modules/postgresql/main.tf
+++ b/modules/postgresql/main.tf
@@ -41,8 +41,14 @@ resource "google_sql_database_instance" "default" {
     disk_size       = "${var.disk_size}"
     disk_type       = "${var.disk_type}"
     pricing_plan    = "${var.pricing_plan}"
-    user_labels     = "${var.user_labels}"
     database_flags  = ["${var.database_flags}"]
+
+    // Define a label to force a dependency to the creation of the network peering.
+    // Substitute this with a module dependency once the module is migrated to
+    // Terraform 0.12
+    user_labels = "${merge(
+      map("tf_dependency", var.peering_completed),
+      var.user_labels)}"
 
     location_preference {
       zone = "${var.region}-${var.zone}"
--- a/modules/postgresql/outputs.tf
+++ b/modules/postgresql/outputs.tf
@@ -25,6 +25,11 @@ output "instance_address" {
   description = "The IPv4 addesses assigned for the master instance"
 }
 
+output private_address {
+  value       = "${google_sql_database_instance.default.private_ip_address}"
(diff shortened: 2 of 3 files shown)


modules/postgresql/main.tf
```diff
@@ -41,9 +41,15 @@ resource "google_sql_database_instance" "default" {
...
     pricing_plan    = "${var.pricing_plan}"
-    user_labels     = "${var.user_labels}"
     database_flags  = ["${var.database_flags}"]
 
+    // Define a label to force a dependency to the creation of the network peering.
+    // Substitute this with a module dependency once the module is migrated to
+    // Terraform 0.12
+    user_labels = "${merge(
+      map("tf_dependency", var.peering_completed),
+      var.user_labels)}"
```
namusyaka: I don't want to add this labels by default. If `peering_completed` is not specified, then `user_labels` should be empty map.

dpetersen: That's understandable. I copied-and-pasted this technique [from the `safer_mysql` submodule](https://github.com/GoogleCloudPlatform/terraform-google-sql-db/blob/master/modules/safer_mysql/main.tf#L36-L41) in this repo, so I think it suffers from the same undesirable behavior.

I am still somewhat new to Terraform, but I'm not sure I can accomplish what you want. I made two attempts at conditionally setting the label ([here](https://github.com/g5search/terraform-google-sql-db/commit/c70175fd880eb90215441d1d40722e6ec5c1d747) and [here](https://github.com/g5search/terraform-google-sql-db/commit/42cdab4bb82f2966e6df0e993a640a99fdf26333)), and both error out with:

```
conditional operator cannot be used with map values
```

Is there an example anywhere else that accomplishes this, so I can use that as a guide? I completely understand why you don't want a `tf_dependency: false` label on CloudSQL instances that aren't using this feature.

namusyaka: How about this?

```hcl
locals {
  peering_completed_enabled = "${var.peering_completed != "" ? true : false}"
  user_labels_including_tf_dependency = {
    enabled = "${merge(map("tf_dependency", var.peering_completed), var.user_labels)}"
    disabled = "${var.user_labels}"
  }
}

resource "google_sql_database_instance" "default" {
  settings {
    user_labels = "${local.user_labels_including_tf_dependency["${local.peering_completed_enabled ? "enabled" : "disabled"}"]}"
  }
}
```

I'm not sure if this works correctly.
If you accept my proposal, please verify two points.

- Does this change work as expected?
- This `tf_dependency` mechanism is to ensure that this module will be run after the VPC network is created. Does this change keep this behavior?

aodj:
How is this PR looking? It's exactly what I'm waiting for (along with 0.12 changes) but it seems to have stalled

dpetersen:
@aodj thank you for the reminder. I have been busy and let this slide. I just made the modifications that @namusyaka suggested and they seem to work as desired. Instances that aren't passed `peering_completed` no longer set the label, while instances with `peering_completed` successfully wait for the peering to be created and are labeled.

Thanks, everybody. Sorry for the delay.

aodj:
@dpetersen Amazing!

namusyaka:
I'm sorry for the late response.
I'll review this change in a few days.
Thanks

morgante:
I'm fine with directly adding it.

dpetersen:
I only really read the MySQL code enough to know what changes needed to be made to support private networking. I actually though the "safe" part was a reference to [`mysqld-safe`](https://dev.mysql.com/doc/refman/8.0/en/mysqld-safe.html)! But now I see it's a hardened alternative to the main module.

While using private networking on CloudSQL Postgres is probably safer than public, there are other reasons to prefer private networking: no need to configure a CloudSQL Proxy and lower latency from your VPC. The CloudSQL proxy feature was our primary motivation to run private instances, since the proxy as a sidecar container [doesn't play well with Kubernetes Jobs](https://github.com/kubernetes/kubernetes/issues/25908). I think you'll want this change in this module regardless of whether you create a hardened `safer_postgresql` module, too.

namusyaka:
Hmm OK. After migrating our modules into 0.12, we should be able to drop the change.

PR #46, opened 2019-07-12
avoid name conflicts when replicas recreation

### LINKED ISSUE #44: Name conflict when recreating failover database
### Scenario ###
A MySQL database restore is done on a database with an existing failover replica, requiring deletion of the failover to complete the restore. Attempting to recreate the failover database using Terraform results in a name collision error.

### Expected Behavior ###
Terraform apply should determine the replica is not present and successfully create it.

### Actual Behavior ###
Name collision error message as below:

`
	* google_sql_database_instance.failover-replica: Error, failed to create instance test-investigate-instance-failover with error code 409: googleapi: Error 409: The Cloud SQL instance already exists., instanceAlreadyExists. This may be due to a name collision - SQL instance names cannot be reused within a week.
`

### Steps to Reproduce ###
1. Using terraform, create a MySQL database with a failover replica
2. Manually restore the database which requires deletion of the failover replica.
3. Run `terraform apply` to recreate the failover replica

### Proposed Solution ###
Create a failover database name suffix that defaults to an empty string to allow the issue above to be resolved and not affect any existing code.

In `variables.tf` add the following code:
```
variable "failover_replica_name_suffix" {
  description = "An optional suffix to add to the failover database name"
  default     = ""
}
```

In 'failover_replica.tf' (line 20):
```
name                  = "${var.name}-failover${(var.failover_replica_name_suffix=="" ? "" : "-")}${var.failover_replica_name_suffix}"
```
The conditional expression in the name above gracefully handles inserting a dash before the suffix if it is provided.

The suffix can now be updated after step 2 above resulting in successful execution of `terraform apply` in step 3.

namusyaka:
To do this, this commit allows you to set the suffix for replicas.
Fixes #44

@morgante @aaron-lane  @wadefrancis could you take a look?

### DIFF
--- a/modules/mysql/failover_replica.tf
+++ b/modules/mysql/failover_replica.tf
@@ -17,7 +17,7 @@
 resource "google_sql_database_instance" "failover-replica" {
   count                 = "${var.failover_replica ? 1 : 0}"
   project               = "${var.project_id}"
-  name                  = "${var.name}-failover${var.failover_replica_name_suffix=="" ? "" : "-"}${var.failover_replica_name_suffix}"
+  name                  = "${var.name}-failover${var.failover_replica_name_suffix}"
   database_version      = "${var.database_version}"
   region                = "${var.region}"
   master_instance_name  = "${google_sql_database_instance.default.name}"
--- a/modules/mysql/failover_replica.tf
+++ b/modules/mysql/failover_replica.tf
@@ -17,7 +17,7 @@
 resource "google_sql_database_instance" "failover-replica" {
   count                 = "${var.failover_replica ? 1 : 0}"
   project               = "${var.project_id}"
-  name                  = "${var.name}-failover"
+  name                  = "${var.name}-failover${var.failover_replica_name_suffix=="" ? "" : "-"}${var.failover_replica_name_suffix}"
   database_version      = "${var.database_version}"
   region                = "${var.region}"
   master_instance_name  = "${google_sql_database_instance.default.name}"
--- a/modules/postgresql/read_replica.tf
+++ b/modules/postgresql/read_replica.tf
@@ -32,7 +32,7 @@ locals {
 resource "google_sql_database_instance" "replicas" {
   count                 = "${var.read_replica_size}"
   project               = "${var.project_id}"
-  name                  = "${var.name}-replica${count.index}${var.read_replica_name_suffix=="" ? "" : "-"}${var.read_replica_name_suffix}"
+  name                  = "${var.name}-replica${var.read_replica_name_suffix}${count.index}"
   database_version      = "${var.database_version}"
   region                = "${var.region}"
   master_instance_name  = "${google_sql_database_instance.default.name}"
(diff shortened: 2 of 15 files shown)


modules/mysql/failover_replica.tf
```diff
@@ -17,7 +17,7 @@
 resource "google_sql_database_instance" "failover-replica" {
   count                 = "${var.failover_replica ? 1 : 0}"
   project               = "${var.project_id}"
-  name                  = "${var.name}-failover"
+  name                  = "${var.name}-failover${var.failover_replica_name_suffix=="" ? "" : "-"}${var.failover_replica_name_suffix}"
```
morgante: Could we just use failover_replica_name_suffix without prepending the hyphen?

wadefrancis: Either way works - I had the bit prepending the hyphen to approximate the way the name had been generated previously to then produce names of the form:
`dbname-failover-suffix`

Simplifying the code to be:
```
name                  = "${var.name}-failover${var.failover_replica_name_suffix}"
```
would result in a name like:
`dbname-failoversuffix`

If a hyphen is then really wanted/needed it can be supplied as part of the suffix itself.

namusyaka: I was thinking prepending hyphen is good for consistency, which has hyphen as delimiters for each word. But I don't have any strong opinions on this, so removed: https://github.com/GoogleCloudPlatform/terraform-google-sql-db/pull/46/commits/a5765c96e3473e4a8a1db09723852abbe0af0d5c

modules/postgresql/read_replica.tf
```diff
@@ -32,7 +32,7 @@ locals {
 resource "google_sql_database_instance" "replicas" {
   count                 = "${var.read_replica_size}"
   project               = "${var.project_id}"
-  name                  = "${var.name}-replica${count.index}"
+  name                  = "${var.name}-replica${var.read_replica_name_suffix}${count.index}"
```
namusyaka: I would like to put `read_replica_name_suffix` before `count.index`. Thoughts?

wadefrancis: Good pickup - that would be another place where this issue would present itself. Ordering also looks good - it would be odd the other way around with a numeral in the middle of the read replica name

morgante: Agreed.

test/fixtures/mysql-ha/main.tf
```diff
@@ -60,6 +60,7 @@ module "mysql" {
   }
 
   // Read replica configurations
+  read_replica_name_suffix                     = "test"
```
morgante: With the change, let's inject the hyphen here.

namusyaka: Done: https://github.com/GoogleCloudPlatform/terraform-google-sql-db/pull/46/commits/173a76b138bd113c9c70e2537b2a09aa17b80973

PR #47, opened 2019-07-12
fix: using network's name in private-service-access module

### LINKED ISSUE #45: Error using module.private-service-access based on the example
Hi,

In the [example](https://github.com/GoogleCloudPlatform/terraform-google-sql-db/blob/1.1.2/examples/mysql-and-postgres/main.tf#L114) for the `module.private_service_access` in v1.1.2, you are using the network's `self_link` property which results in an error : 

```sh
* google_compute_global_address.google-managed-services-range: Error creating GlobalAddress: googleapi: Error 400: Invalid value for field 'resource.name': 'google-managed-services-https://www.googleapis.com/compute/v1/projects/my-project/global/networks/main'. Must be a match of regex '(?:[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?)', invalid
```

If I am not mistaken the [module documentation](https://github.com/GoogleCloudPlatform/terraform-google-sql-db/tree/master/modules/private_service_access) states : `vpc_network | Name of the VPC network to peer.`

So it should be :

```diff
  module "private-service-access" {
    source      = "../../modules/private_service_access"
    project_id  = "${var.project_id}"
-   vpc_network = "${google_compute_network.default.self_link}"
+   vpc_network = "${google_compute_network.default.name}"
  }
```

Once fixed, the module works as intended.

BastienM:
Fixes #45

### DIFF
--- a/examples/mysql-and-postgres/main.tf
+++ b/examples/mysql-and-postgres/main.tf
@@ -114,7 +114,7 @@ module "postgresql-db" {
 module "private-service-access" {
   source      = "../../modules/private_service_access"
   project_id  = "${var.project_id}"
-  vpc_network = "${google_compute_network.default.self_link}"
+  vpc_network = "${google_compute_network.default.name}"
 }
 
 module "safer-mysql-db" {


BastienM:
CLA signed.

morgante:
In theory, self-link and name should be mostly interchangeable. In practice, this isn't always the case so thank you for the contribution.

PR #56, opened 2019-08-16
deprecate root module

namusyaka:
As we haven't maintained the root module for a while, it's time to deprecate the module.
@morgante @danisla  What do you think?


morgante:
@namusyaka I agree

@danisla Can you approve as well?

danisla:
LGTM

PR #57, opened 2019-08-16
fix a few things

namusyaka:
I have found a few points that should be fixed.


modules/postgresql/outputs.tf
```diff
@@ -20,12 +20,12 @@ output "instance_name" {
   description = "The instance name for the master instance"
 }
 
-output "instance_address" {
+output "instance_ip_address" {
```
namusyaka: This change is reasonable? @morgante

morgante: I think we should leave it as is.

namusyaka: OK.
We'll be able to deprecate this and rename in the future (probably next major version bump?)

namusyaka: Fixed: https://github.com/terraform-google-modules/terraform-google-sql-db/pull/57/commits/865b976daab04259fa953ee9de04bd2d9e156c3a
Issue opened: #58

Issue #59, opened 2019-09-23
Replace extraneous label with module dependency in PostgreSQL submodule


aaron-lane:
This issue is extracted from #43.

aaron-lane:
https://github.com/terraform-google-modules/terraform-google-sql-db/blob/master/modules/postgresql/main.tf#L82-L84

aaron-lane:
https://github.com/terraform-google-modules/terraform-google-sql-db/blob/master/modules/safer_mysql/main.tf#L36-L38

aaron-lane:
Blocked by https://github.com/hashicorp/terraform/issues/10462

aaron-lane:
We can work around the preceding issue by using a strategy like the Network module: https://github.com/terraform-google-modules/terraform-google-network/blob/master/modules/network-peering/variables.tf#L45

So to clarify, the `peering_completed` variable and `tf_dependency` label should be removed, and a `module_depends_on` variable should be added and used to create a dependency for the `google_sql_database_instance` resources.

PR #77, opened 2019-12-03
[postgresql] Add module dependency for peering network.

omazin:
Fixes https://github.com/terraform-google-modules/terraform-google-sql-db/issues/59


aaron-lane:
#59 is currently blocked.

PR #98, opened 2020-03-26
complete provision instructions

PascalBourdier:
cause this is not enough to use this module from the public registry

```
module "sql-db" {
  source  = "GoogleCloudPlatform/sql-db/google"
  version = "3.1.0"
}
```

This warn help us, but can be more explicit :
This module version (3.1.0) has no root configuration. A module with no root configuration cannot be used directly.
Use the submodules dropdown above to view the 4 submodules defined within this module.

So add a small usage guideline in the readme and examples with remote usage can help us


PascalBourdier:
> Examples are used for tests and should not be modified.

so I only change the README now

Issue #71, opened 2019-11-19
Enabling high availability for mysql instances


alejandro-delgado-garcia:
Is there any support in this module to enable high availability for MySQL instances?

morgante:
Yes, you can configure the failover replica and associated settings. See the available inputs here: https://github.com/terraform-google-modules/terraform-google-sql-db/tree/master/modules/mysql#inputs

alejandro-delgado-garcia:
Hi @morgante it seems that cloud SQL 2nd generation instances doesn't use a replica for the failover according to google docs https://cloud.google.com/sql/docs/mysql/high-availability#legacy_mysql_high_availability_option

alejandro-delgado-garcia:
In the Terraform docs, there is a property not supported by this module https://www.terraform.io/docs/providers/google/r/sql_database_instance.html#availability_type which might work to enable this, although in the docs they just mentioned Postgres but I think it could also work for MySQL

morgante:
I see, yes it looks like we would have to do some work to support the new version of high availability.

axot:
Hi, due to the official document(https://cloud.google.com/sql/docs/mysql/configure-ha), I think it should work to setup with `availabilityType":"REGIONAL"` and `backupConfiguration` in current version.

```
curl --header "Authorization: Bearer ${ACCESS_TOKEN}" \
     --header 'Content-Type: application/json' \
     --data '{"name":"[INSTANCE_NAME]", "region":"[REGION]",
              "databaseVersion":"[DATABASE_VERSION]",
              "settings": {
                  "tier":"[MACHINE_TYPE]",
                  "backupConfiguration": {"binaryLogEnabled": true, "enabled": true},
                  "availabilityType":"REGIONAL"}}' \
     -X POST \
     https://www.googleapis.com/sql/v1beta4/projects/[PROJECT-ID]/instances
```

alejandro-delgado-garcia:
Hi everyone, is there any progress about this issue? Thanks!

morgante:
@alejandro-delgado-garcia Not currently, though I'd be happy to review a PR if anyone wants to take a stab at it.

mehdicopter:
It works the same way as postgres.
Just add `availability_type = var.availability_type` into `settings` and add this block into variables.tf and it will work like a charm on MySQL instances.

```
variable "availability_type" {
  description = "The availability type for the master instance. Can be either ZONAL or REGIONAL."
  type        = string
[...]

Issue #97, opened 2020-03-24
Remove deprecated failover instance for HA MySQL


Dev25:
The usage of a failover instance for mysql has been deprecated in favour of the newer `availability-type=REGIONAL` method that Postgres is already using.

https://cloud.google.com/sql/docs/mysql/high-availability#legacy_mysql_high_availability_option

>Until April 2020, you have the option of using the legacy process for adding high availability to MySQL instances, which uses a failover replica. The legacy functionality is not available in the Cloud Console. Instead, use gcloud or cURL commands. See Legacy configuration: Creating a new instance configured for high availability or Legacy configuration: Configuring an existing instance for high availability.

Related Issue on using the new HA field: https://github.com/terraform-google-modules/terraform-google-sql-db/issues/71

PR #99, opened 2020-03-30
Support new regional HA MySQL

Dev25:
Closes https://github.com/terraform-google-modules/terraform-google-sql-db/issues/71

https://cloud.google.com/sql/docs/mysql/high-availability#HA-configuration
>The HA configuration, sometimes called a cluster, provides data redundancy. A Cloud SQL instance configured for HA is also called a regional instance and is located in a primary and secondary zone within the configured region. Within a regional instance, the configuration is made up of a primary instance (master) and a standby instance. Through synchronous replication to each zone's persistent disk, all writes made to the primary instance are also made to the standby instance. In the event of an instance or zone failure, this configuration reduces downtime, and your data continues to be available to client applications.

>Note: The standby instance cannot be used for read queries. This differs from the Cloud SQL for MySQL legacy HA configuration.

WIP still need to test/update docs

I can also remove the legacy replica stuff for https://github.com/terraform-google-modules/terraform-google-sql-db/issues/97 but that does mean a breaking release/2 step migration path ([Updating an instance from legacy to current high availability](https://cloud.google.com/sql/docs/mysql/configure-ha#updating_an_instance_from_legacy_to_current)). Thoughts @morgante ?

### DIFF
--- a/modules/mysql/main.tf
+++ b/modules/mysql/main.tf
@@ -22,6 +22,10 @@ locals {
     enabled  = var.ip_configuration
     disabled = {}
   }
+
+  // HA method using REGIONAL availability_type requires binary logs to be enabled
+  binary_log_enabled = var.availability_type == "REGIONAL" ? true : lookup(var.backup_configuration, "binary_log_enabled", null)
+  backups_enabled    = var.availability_type == "REGIONAL" ? true : lookup(var.backup_configuration, "enabled", null)
 }
 
 resource "google_sql_database_instance" "default" {
@@ -33,12 +37,13 @@ resource "google_sql_database_instance" "default" {
   settings {
     tier                        = var.tier
     activation_policy           = var.activation_policy
+    availability_type           = var.availability_type
     authorized_gae_applications = var.authorized_gae_applications
     dynamic "backup_configuration" {
       for_each = [var.backup_configuration]
       content {
-        binary_log_enabled = lookup(backup_configuration.value, "binary_log_enabled", null)
-        enabled            = lookup(backup_configuration.value, "enabled", null)
+        binary_log_enabled = local.binary_log_enabled
+        enabled            = local.backups_enabled
         start_time         = lookup(backup_configuration.value, "start_time", null)
       }
     }
--- a/modules/mysql/variables.tf
+++ b/modules/mysql/variables.tf
@@ -56,9 +56,9 @@ variable "activation_policy" {
 }
 
 variable "availability_type" {
-  description = "The availability type for the master instance. Can be either `ZONAL` or `REGIONAL`."
+  description = "The availability type for the master instance. Can be either `REGIONAL` or `null`."
   type        = string
-  default     = "ZONAL"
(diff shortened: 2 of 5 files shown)


craigafinch:
It would be very helpful if someone could expedite review for this feature. We are unable to use Terraform to manage Cloud SQL instances with high availability until this feature is released.

modules/mysql/variables.tf
```diff
@@ -55,6 +55,12 @@ variable "activation_policy" {
   default     = "ALWAYS"
 }
 
+variable "availability_type" {
+  description = "The availability type for the master instance. Can be either `ZONAL` or `REGIONAL`."
+  type        = string
+  default     = "ZONAL"
```
wwwsevolod, line 61: I recommend to set default to REGIONAL as it is new default value and in July it will be even not possible to change it to ZONAL, also, also this info probably worth adding to description

rafael-romero-carmona-claranet: Either REGIONAL or null but not ZONAL. Both values (regional and null) will be correct in the future as well.

Dev25: Fixed in c247027 by setting to `REGIONAL`, description also updated to reflect this.

> in July it will be even not possible to change it to ZONAL

It's not going to be possible to create non HA masters in CloudSQL?

rafael-romero-carmona-claranet: I think that null is the value to create non-ha CloudSQL

modules/mysql/main.tf
```diff
@@ -22,6 +22,10 @@ locals {
     enabled  = var.ip_configuration
     disabled = {}
   }
+
+  // HA method using REGIONAL availability_type requires binary logs to be enabled
```
aharelick, line 26: Is it true that `REGIONAL` `availability_type` requires binary logs? From poking around the docs it seems like binary logs are only required if you want to use a read replica or if you want point in time recovery?

>These settings determine whether automated backups are performed and whether binary logging is enabled or not. Both options add a small performance cost, but are required for the creation of replicas and clones, and for point-in-time recovery.

https://cloud.google.com/sql/docs/mysql/instance-settings#backups-and-binary-logging-2ndgen

Dev25: https://github.com/terraform-google-modules/terraform-google-sql-db/issues/71#issuecomment-604998264

It fails without bin logs enabled, the GCP docs mention this for enabling HA

https://cloud.google.com/sql/docs/mysql/configure-ha#console
https://cloud.google.com/sql/docs/mysql/configure-ha#gcloud

>Leave the Automate backups and Enable point-in-time recovery checkboxes enabled.

Which requires bin logs

aharelick: Very interesting, thanks for those links. I've recreated the same thing for myself with gcloud, the REST API, and the GCP UI.

I'll follow up with a GCP Support ticket to understand if they'll offer more flexibility in the future. Thanks for writing up the PR!

PR #100, opened 2020-04-02
Move users/databases to for_each

### LINKED ISSUE #96: Move replica/users/databases to for_each
`count`  is being used for 
- Databases `var.additional_databases`
- Users  `var.additional_users`
- Replica instances `var.read_replica_size`

We should move that to using `for_each` keeping the same input type (list) and just constructing the map/set internally similar to other modules.

Dev25:
^
Closes #96

Noticed a couple of object fields `instance` and` project` were never used so also removed them

To consider
- Merge `users` and `additonal_users`  + same for databases, seems a bit pointless having them seperate and since we've already done a breaking API change might as well merge them now.
- Move read_replicas to `for_each`

### DIFF
--- a/examples/mysql-ha/main.tf
+++ b/examples/mysql-ha/main.tf
@@ -190,8 +190,6 @@ module "mysql" {
 
   additional_databases = [
     {
-      project   = var.project_id
-      instance  = local.instance_name
       name      = "${var.mysql_ha_name}-additional"
       charset   = "utf8mb4"
       collation = "utf8mb4_general_ci"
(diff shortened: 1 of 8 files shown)


Dev25:
This closes #100

bharathkkb:
Overall LGTM
Regarding points to consider, I havent used this module much but
Arent `read_replicas` defined under `/modules/[mysql/pgsql]/read_replica.tf` already using `for_each`

Dev25:
> Arent `read_replicas` defined under `/modules/[mysql/pgsql]/read_replica.tf` already using `for_each`

The replica configuration blocks are done via `for_each` but not the instance resource

```tf
resource "google_sql_database_instance" "replicas" {
  count                = var.read_replica_size
```

Dev25:
I think for now lets ignore read replicas, imo that entire section could be more flexible e.g. different instance tiers/settings for different replicas so if/whenever that changes we could move to for_each then as well.

syedrakib:
Do we have any ETA on this?

bharathkkb:
@Dev25 we have a conflict, can you rebase ?
@morgante PTAL

PR #116, opened 2020-05-22
Added: random_id for mysql instance name

konstantin-recurly:
There is an issue if you would like to create a DB with the same name. With the current approach, you need to change the name every time or you have to work around and use something like `random_id`

> **If I delete my instance, can I reuse the instance name?**
> Yes, but not right away. The instance name is unavailable for up to a week before it can be reused.

More information is here [Managing Your Instances](https://cloud.google.com/sql/faq?hl=en#managing-your-instances)

### DIFF
--- a/modules/mysql/main.tf
+++ b/modules/mysql/main.tf
@@ -38,7 +38,7 @@ resource "random_id" "prefix" {
 resource "google_sql_database_instance" "default" {
   provider            = google-beta
   project             = var.project_id
-  name                = "${var.name}-${random_id.prefix.hex}"
+  name                = var.random_instance_name ? "${var.name}-${random_id.prefix.hex}" : var.name
   database_version    = var.database_version
   region              = var.region
   encryption_key_name = var.encryption_key_name
--- a/modules/mysql/main.tf
+++ b/modules/mysql/main.tf
@@ -31,14 +31,16 @@ locals {
   backups_enabled    = var.availability_type == "REGIONAL" ? true : lookup(var.backup_configuration, "enabled", null)
 }
 
-resource "random_id" "prefix" {
+resource "random_id" "suffix" {
+  count = var.random_instance_name ? 1 : 0
+
   byte_length = 4
 }
 
 resource "google_sql_database_instance" "default" {
   provider            = google-beta
   project             = var.project_id
-  name                = var.random_instance_name ? "${var.name}-${random_id.prefix.hex}" : var.name
+  name                = var.random_instance_name ? "${var.name}-${random_id.suffix.hex}" : var.name
   database_version    = var.database_version
   region              = var.region
   encryption_key_name = var.encryption_key_name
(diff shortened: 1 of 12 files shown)


morgante:
I'm not sure we want to include this in the module. While it's indeed necessary for recreating instances after deletion, it's also following a very specific pattern which doesn't necessarily match what everyone wants.

konstantin-recurly:
Maybe, it would be nice to have that as an option. Like
```hcl
resource "random_id" "prefix" {
  byte_length = 4
}

resource "google_sql_database_instance" "default" {
  provider            = google-beta
  project             = var.project_id
  name                = var.instance_name_prefix  ? "${var.name}-${random_id.prefix.hex}"  : var.name
```
How does it sound to you?

morgante:
Yeah I think we could make it optional with a `random_instance_name` boolean variable.

konstantin-recurly:
@morgante Updated the PR

modules/mysql/variables.tf
```diff
@@ -24,6 +24,12 @@ variable "name" {
   description = "The name of the Cloud SQL resources"
 }
 
+variable "random_instance_name" {
+  type        = bool
+  description = "Sets random prefix at the end of the Cloud SQL resource name"
```
morgante, line 29:
```suggestion
  description = "Sets random suffix at the end of the Cloud SQL resource name"
```

modules/mysql/main.tf
```diff
@@ -31,10 +31,14 @@ locals {
   backups_enabled    = var.availability_type == "REGIONAL" ? true : lookup(var.backup_configuration, "enabled", null)
 }
 
+resource "random_id" "prefix" {
```
morgante, line 34: We should only generate this if we need it:
```
count = var.random_instance_name ? 1 : 0
```

modules/mysql/variables.tf
konstantin-recurly: oh, right

modules/mysql/main.tf
konstantin-recurly: added `count`

konstantin-recurly:
One more thing that I have noticed. Seems that now we don't need this part in examples
https://github.com/terraform-google-modules/terraform-google-sql-db/blob/master/examples/mysql-public/main.tf#L37-L44
```hcl
resource "random_id" "suffix" {
  byte_length = 5
}

locals {
  /*
    Random instance name needed because:
    "You cannot reuse an instance name for up to a week after you have deleted an instance."
    See https://cloud.google.com/sql/docs/mysql/delete-instance for details.
  */
  instance_name = "${var.db_name}-${random_id.suffix.hex}"
}
```
Now we should be able to use this
```hcl
module "safer-mysql-db" {
  source           = "../../modules/mysql"
  name             = var.db_name
  random_instance_name = true
  database_version = "MYSQL_5_6"
  project_id = var.project_id
```

morgante:
Indeed! Looks like that example could be cleaned up.

We can also remove random_Id from safer-mysql and just pass through the `random_instance_name` variable: https://github.com/terraform-google-modules/terraform-google-sql-db/blob/master/modules/safer_mysql/main.tf#L16

Do you mind doing a follow-up PR to fix those issues?

PR #117, opened 2020-05-27
Replica Rework

Dev25:
This PR is a breaking change and cleans up the failover + replicas configs. Lots of config options that only worked for 1st gen instances or had no affect at all have been removed.

Failover instance has been completely removed, 3.2 release moved to using new regional HA instances.

#### Replica changes
- Move to using for_each + list input allowing different replica values per replica
- Remove a bunch of vars that only applied to first generation instances e.g. `crash_safe_replication`, `replication_type`
- Hardcode some vars that only have 1 possible option e.g. `pricing_plan` only has 1 valid option, read replicas can't be shutdown so `activation_policy` is `ALWAYS`
- Simplify `replica_configuration` 
- Allow cross regional replicas! (New Feature)
- Remove `maintenance_window` which isn't applicable to replicas. ~This breaks the migration path because you can't change the values once created~ Actually `ignore_changes` should be able to provide a migration path.

##### Example Usage
```terraform
  read_replica_name_suffix = "-"
  read_replicas = [
    {
      name   = "c"
      tier   = "db-n1-standard-1"
      region = null
      zone   = "europe-west1-c"
      ip_configuration = {
        ipv4_enabled    = true
        require_ssl     = true
        private_network = var.host_vpc_link
        authorized_networks = []
      }

      database_flags  = var.mysql_database_flags
      disk_autoresize = null  # Can probably force this to master var?
      disk_size       = null  # Can probably force this to master var?
      disk_type       = null
      user_labels = {
        env = "production"
      }
    },
  ]

```

This still requires a bit of cleanup + examples/tests but what are early thoughts on the direction?


morgante:
I'm supportive of the general direction and agree we can phase out failover replica support. The main concerns I would have include: providing a clear upgrade path for existing configs and making sure we roll the changes out to all 3 modules simultaneously.

Dev25:
It looks like its only mysql+postgres support read replicas, i'll update the postgres module after fixing up the MySQL examples/tests first.

morgante:
Thanks, do you have a sense of when you'll be able to complete it? Since this will be a breaking change, I might hold 4.0.0 to include it if you'll be ready in the next ~week.

Dev25:
By end of next week sounds about right

Dev25:
@morgante I still need to test the migration path but you can review API changes in the meantime.

modules/mysql/read_replica.tf
```diff
@@ -80,39 +53,32 @@ resource "google_sql_database_instance" "replicas" {
         }
       }
     }
-    authorized_gae_applications = var.authorized_gae_applications
 
-    crash_safe_replication = var.read_replica_crash_safe_replication
-    disk_autoresize        = var.read_replica_disk_autoresize
```
morgante, line 86: What's the reason for removing this?

modules/mssql/variables.tf
```diff
@@ -24,12 +24,6 @@ variable "name" {
   description = "The name of the Cloud SQL resources"
 }
 
-variable "random_instance_name" {
```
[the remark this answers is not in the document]
Dev25: You're right, rebasing/updating from master screwed that up, will fix.

modules/mysql/read_replica.tf
Dev25: re: `crash_safe_replication` its leftovers from the now removed 1st gen instances. 
From Provider docs:
> Remove any fields that are not applicable to Second-generation instances: * settings.crash_safe_replication * settings.replication_type * settings.authorized_gae_applications

https://cloud.google.com/sql/docs/mysql/admin-api/rest/v1beta4/instances
ctrl+f for `First Generation instances` to see further differences

morgante: Right, I meant the `read_replica_disk_autoresize` variable.

Dev25: That's been removed in favour of using the master instance `var.disk_autoresize` to help simplify things and avoid potential problems down the line of users just enabling it on the master and not replicas. If the master disk increases, then replicas should be the same.

The same approach has been taken for `var.read_replica_disk_size` where its completely removed in favour of `var.disk_size`.

Although i can see a edge case where users may want a replica to have a larger disk size vs master (and they aren't using autoresize/near capacity already) due to certain queries creating temp tables on disk etc. I'm not sure how likely that scenario is/trying to minimise object fields since we still can't default them. Opinion on having that `disk_size` replica option available?

morgante: I'd suggest we leave the existing `read_replica_disk_size` and `read_replica_disk_autoresize` untouched (to cover those corner cases) but default to using the master's settings.

modules/mssql/variables.tf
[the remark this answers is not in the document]
Dev25: Fixed via https://github.com/terraform-google-modules/terraform-google-sql-db/pull/117/commits/56c41e4367bca3b37a534255039be537ff791312 + refactored into local var https://github.com/terraform-google-modules/terraform-google-sql-db/pull/117/commits/9bb30b457fc8dad633cb5e84f34bd4c7b8ab5e1a

modules/mysql/read_replica.tf
Dev25: Fixed via https://github.com/terraform-google-modules/terraform-google-sql-db/pull/117/commits/8c97fe90475652cae7b9f5607048167737a0a2fe

Dev25:
I've tested the migration path for a mysql instance from v3.2 and it LGTM, just a simple state mv + move replica config to new var.

PR #125, opened 2020-07-06
Set location for backup_configuration

syedrakib:
Add ability to set "location" in mysql `backup_configuration`

### DIFF
--- a/examples/mysql-ha/main.tf
+++ b/examples/mysql-ha/main.tf
@@ -80,6 +80,7 @@ module "mysql" {
     enabled            = true
     binary_log_enabled = true
     start_time         = "20:55"
+    location           = "us"
   }
 
   // Read replica configurations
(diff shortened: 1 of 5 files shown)


bharathkkb:
@syedrakib thanks for the PR! Could you modify the examples where `backup_configuration ` is used like https://github.com/terraform-google-modules/terraform-google-sql-db/blob/295ca1842b4bd4121b0618fedf773c7e5363f40a/examples/mysql-ha/main.tf#L79
You can also run the lint tests locally by running `make docker_test_lint`. More details [here](https://github.com/terraform-google-modules/terraform-google-sql-db/blob/master/CONTRIBUTING.md).

syedrakib:
Thanks for the advice @bharathkkb . I will fix these matters.

morgante:
Fixed in #126

PR #135, opened 2020-08-11
Update versions.tf

cocktailer:


modules/private_service_access/versions.tf
```diff
@@ -15,7 +15,7 @@
  */
 
 terraform {
-  required_version = "~> 0.12.6"
+  required_version = ">= 0.13"
```
bharathkkb, line 18:
```suggestion
  required_version = ">=0.12.6, <0.14"
```

mpastecki:
Would it be possible to release a tag with this before [4.0.0](https://github.com/terraform-google-modules/terraform-google-sql-db/pull/110) is released?

PR #149, opened 2020-10-06
added note in document for disk size variable

sumeetoc:


morgante (approved):
Thanks for adding the note.

PR #152, opened 2020-10-23
Feature/add deletion protection option

askhari:
Added a _deletetion_protection_ variable to all modules. Default value: _true_.

This will allow a more easy way to delete managed databases. The default value is the same I found in the _terraform-provider-google_ to follow the same behaviour.


askhari:
I see that there is an error in the integration tests, but I cannot see the log of the build in GCP. Could you please let me know what is failing to work on it?

morgante:
Duplicate of #151.

PR #155, opened 2020-11-03
Read replica deletion protection

WillRemesh:
This PR adds a variable to manage read replica deletion protection.


morgante (commented):
I'm not sure we need a separate variable for replicas?

WillRemesh:
I thought it would be helpful, usually replicas aren't as important to keep up. Could set read_replica_deletion_protection to false always, then only change deletion_protection when changing the default db. I'm open to using the existing variable or setting to false always, just let me know the preference and I can make the change. Currently can't bring down replicas with TF so this is blocking for us.

morgante:
Got it, that's fine. Can you please get tests passing, as per the contributing guide?

WillRemesh:
I'm having trouble getting the test setup project working. seeing: 
```
Error: Error parsing /var/lib/jenkins/workspace/dsl_custom_terraform/test/setup/iam.tf: At 25:18: Unknown token: 25:18 IDENT module.project.project_id.
```
I also can't see what the error is in the CI since I don't have permissions. Any help with either more info about what is failing the tests or how to get the test/setup env working?

WillRemesh:
Is it possible to give me permissions so I can see the build failures? I'm not getting any more failures locally and having a hard time debugging this. I had to manually clean my local to get the integration test to work. make docker_test_cleanup did not remove enough of the files to work. Is it possible this is what is breaking the build or is it a fresh env for each build?

WillRemesh:
@morgante Tests are now passing, sorry this took so long, had to fix things from previous merged PR's.

modules/postgresql/variables.tf
```diff
@@ -269,3 +269,9 @@ variable "deletion_protection" {
   type        = bool
   default     = true
 }
+
+variable "read_replica_deletion_protection" {
+  description = "Used to block Terraform from deleting replica SQL Instances."
+  type        = bool
+  default     = true
```
morgante, line 276: I'm wondering if this should be false by default? Deleting read replicas isn't as dangerous as deleting the primary.

WillRemesh: Updated, also removed `read_replica_deletion_protection = false` from the examples

morgante (approved):
Thanks!

PR #161, opened 2020-11-17
Have random name propagate to replicas also

WillRemesh:
Not able to send a random replica name through to replicas since they are using for each now in version 4.*.

This piggybacks on the random_instance_name variable to also append random suffix to replica names. This is needed to be able to bring replica's up and down within 24 hours due to GCP naming conflict limitation. 

Open to creating a new variable for replicas (replica_random_instance_name?), or else this is backwards incompatible for users with existing replicas and random_instance_name set to true.


WillRemesh:
breaking example is here. https://github.com/Remesh/terraform-google-sql-db/pull/4

WillRemesh:
Updated this to it's own variable since it breaks if not currently using random instance name for default db

morgante:
@WillRemesh Do you mind rebasing? Sorry for missing this earlier.

PR #162, opened 2020-11-17
Example of passing random replica suffix to replica name

WillRemesh:
Breaks TF deploy because of TF for_each dependencies.

This is not alleviated with module_depends_on or tf13 depends_on inputs.


WillRemesh:
Wanted to open as example on form

Issue #159, opened 2020-11-16
Allow multiple google_compute_global_address per VPC


odise:
I ran into problems while trying to attach multiple mssql cluster to the same VPC. 

```
Error: Error creating GlobalAddress: googleapi: Error 409: The resource 'projects/********************/global/addresses/google-managed-services-dev-vpc' already exists, alreadyExists

  on .terraform/modules/sql.private_service_access/modules/private_service_access/main.tf line 27, in resource "google_compute_global_address" "google-managed-services-range":
  27: resource "google_compute_global_address" "google-managed-services-range" {
```

#158 should fix this. The `suffix` parameter could be a `random_id` or something else.

morgante:
@odise Can you clarify what you're trying to accomplish? A single private service connection should actually be sufficient for multiple CloudSQL clusters.

odise:
@morgante I would like to add an additional IP range to the private service connection which has been already created by the `private_service_access` module. My previous attempts failed when adding a new mssql instance using the existing one:

```
Failed to create subnetwork. Couldn't find free blocks in allocated IP ranges. Please allocate new ranges for this service provider.
```

Creating an additional `google_compute_global_address` resource should exactly accomplish that. However the name of that resource is already taken (see the error in the description). Maybe the `google_service_networking_connection` resource shouldn't be recreated in that case?

odise:
After investigating quite a bit I found that #158 is not necessary in order to setup multiple Cloud SQL instances attached to the same VPC. Sorry for the inconvenience.

PR #158, opened 2020-11-16
feat: allow multiple google_compute_global_address per vpc

odise:

PR #178, opened 2021-01-05
feat: allow multiple google_compute_global_address per vpc - reloaded

odise:
I open this PR in reference to #158 and #159 from last year. It turned out that private service connections needs to be created for each region and Cloud SQL type peered with the VPC:

> Cloud SQL allocates a /24 subnet from the private services access IP range for each combination of region and database type. For example, placing MySQL instances in two regions requires that the allocated IP address range contain at least two available subnets of size /24, and deploying MySQL and PostgreSQL in two regions requires four available subnets of size /24. SQL Server instances may share the same subnet with MySQL instances in a region.

Please have a look at https://cloud.google.com/sql/docs/mysql/private-ip#network_requirements .

With the current implementation of the `private_service_access` this is not possible as the name of the `google_compute_global_address` resource is generated from the VPC network. I added a suffix to overcome this problem. This way multiple instances of the `private_service_access` can be deployed for a VPC making it possible to deploy multiple Cloud SQL types (e.g. MSSQL and postgresql).

### DIFF
--- a/modules/private_service_access/variables.tf
+++ b/modules/private_service_access/variables.tf
@@ -53,3 +53,4 @@ variable "name_suffix" {
   type        = string
   default     = ""
 }
+
--- a/modules/private_service_access/variables.tf
+++ b/modules/private_service_access/variables.tf
@@ -47,3 +47,9 @@ variable "labels" {
   type        = map(string)
   default     = {}
 }
+
+variable "name_suffix" {
+  description = "An optional suffix added to resource names."
+  type        = string
+  default     = ""
+}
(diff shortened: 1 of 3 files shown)


modules/private_service_access/variables.tf
```diff
@@ -47,3 +47,9 @@ variable "labels" {
   type        = map(string)
   default     = {}
 }
+
+variable "name_suffix" {
+  description = "An optional suffix added to resource names."
+  type        = string
+  default     = ""
+}
```
morgante, line 55: Please add a newline at the end of the file.

PR #189, opened 2021-02-18
Relax random provider

dor961:
Ran into provider dependency issues with other modules in the same project.


modules/postgresql/versions.tf
```diff
@@ -19,6 +19,5 @@ terraform {
   required_providers {
     google = "~> 3.5"
     null   = "~> 2.1"
-    random = "~> 2.2"
```
morgante, line 22: You should at least pin a minimum version.

morgante (changes requested):

modules/postgresql/versions.tf
dor961: Could you please review the latest commit?
Thanks.

morgante (approved):

billyfoss:
Is there any reason the null and random providers cannot use the 3.x version?  The breaking change in 3.x is dropping support for Terraform 0.11, which is not relevant to this module.

morgante:
> Is there any reason the null and random providers cannot use the 3.x version? The breaking change in 3.x is dropping support for Terraform 0.11, which is not relevant to this module.

We should be fine to allow this.

morgante:
This has been fixed.

comment-bot-dev:

PR #190, opened 2021-02-18
Adding conditional flag to enable read replica if required

dhinakar2707:
HI team,

We are using terraform open-source terraform module for creating postgres sql instance...where we are using same code for creating postgres sql instance in multiple environments like dev, staging & production. 
Where we require read replica instance only for production, it's not needed for dev & staging. With the current master code we are not able to restrict this approach, so I have added a count condition with a boolean flag. Please have a review on this PR and let me know if this make sense. Also let me know if i have missed something from my end.

### DIFF
--- a/modules/postgresql/read_replica.tf
+++ b/modules/postgresql/read_replica.tf
@@ -21,6 +21,7 @@ locals {
 }
 
 resource "google_sql_database_instance" "replicas" {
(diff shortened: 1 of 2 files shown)


morgante:
Why do you need a conditional flag? You can already set `read_replicas` to an empty list.

PR #185, opened 2021-01-26
add encryption_key_name to safer_mysql module

rafstef:
add encryption_key_name var to safer_mysql module

PR #193, opened 2021-02-19
Add outputs for the actual `google_sql_database_instance` resources

andreykaipov:
Addresses part of https://github.com/terraform-google-modules/terraform-google-sql-db/issues/192

PR #195, opened 2021-02-19
lint to fix build

andreykaipov:
PR https://github.com/terraform-google-modules/terraform-google-sql-db/pull/185 added trailing whitespace which is breaking builds


andreykaipov:
Sorry for the PR spam. Since I've got both #193 and #194 open I figured it'd be best to open up this one for the linting rather than sneaking it in with one of those two.

morgante (approved):

Issue #188, opened 2021-02-15
Support for Cloud SQL Insights


expobrain:
Now that GCP provides the Cloud SQL Insights tool, will it be supported by this module in the near future?

morgante:
Please request it in the provider first: https://github.com/hashicorp/terraform-provider-google

Tiduster:
Insights is now available in the provider :-) .

PR #198, opened 2021-03-03
Support Query Insights

dhoeric:
For #188 since provider already supporting insights config (https://github.com/hashicorp/terraform-provider-google/pull/8434).

### DIFF
--- a/modules/postgresql/main.tf
+++ b/modules/postgresql/main.tf
@@ -24,6 +24,13 @@ locals {
     disabled = {}
   }
 
+  insights_config_enabled = lookup(var.insights_config, "query_insights_enabled", false)
+
+  insights_config = {
+    enabled  = [var.insights_config]
+    disabled = []
+  }
+
   databases = { for db in var.additional_databases : db.name => db }
   users     = { for u in var.additional_users : u.name => u }
 }
@@ -76,7 +83,7 @@ resource "google_sql_database_instance" "default" {
       }
     }
     dynamic "insights_config" {
-      for_each = [var.insights_config]
+      for_each = local.insights_config[local.insights_config_enabled ? "enabled" : "disabled"]
       content {
         query_insights_enabled  = lookup(insights_config.value, "query_insights_enabled", false)
         query_string_length     = lookup(insights_config.value, "query_string_length", 1024)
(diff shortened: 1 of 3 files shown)


modules/postgresql/variables.tf
```diff
@@ -138,6 +138,22 @@ variable "backup_configuration" {
   }
 }
 
+variable "insights_config" {
+  description = "The insights_config settings subblock for the database setings"
+  type = object({
+    query_insights_enabled  = bool
```
morgante, line 144: Instead of having this, we should default to `insights_config` being null and set `query_insights_enabled = true` automatically if the block is provided.

morgante (changes requested):

modules/postgresql/variables.tf
dhoeric: Thanks for the review, I have patched for the advice.

morgante: Sorry, I think I might have been unclear. I want the default for this entire variable to be `null` (disable query insights entirely). Like so:

```suggestion
    query_string_length     = number
    record_application_tags = bool
    record_client_address   = bool
  })
  default = null
```

morgante (changes requested):

modules/postgresql/main.tf
```diff
@@ -75,6 +87,15 @@ resource "google_sql_database_instance" "default" {
         }
       }
     }
+    dynamic "insights_config" {
+      for_each = local.insights_config[var.insights_config != null ? "enabled" : "disabled"]
```
morgante, line 91: This logic seems a little confusing. Can't we do it like this?

```suggestion
      for_each = var.insights_config != null ? [var.insights_config] : []
```

morgante (changes requested):

modules/postgresql/variables.tf
```diff
@@ -138,6 +138,16 @@ variable "backup_configuration" {
   }
 }
 
+variable "insights_config" {
+  description = "The insights_config settings subblock for the database setings"
```
morgante, line 142:
```suggestion
  description = "The insights_config settings for the database."
```

morgante:
@dhoeric  You need to regenerate docs.

morgante (approved):

PR #208, opened 2021-04-22
add region tags to public instance creation examples to embed in publ…

tlichtenberg:
add region tags to public instance creation examples to embed in public Cloud SQL documentation

### DIFF
--- a/examples/mssql-public/main.tf
+++ b/examples/mssql-public/main.tf
@@ -19,6 +19,7 @@ provider "google-beta" {
   region  = var.region
 }
 
+# [START cloud_sql_sqlserver_instance_create_public]
 module "mssql" {
   source               = "../../modules/mssql"
   name                 = var.name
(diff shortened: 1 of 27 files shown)


comment-bot-dev:

examples/mssql-public/main.tf
```diff
@@ -19,6 +19,7 @@ provider "google-beta" {
   region  = var.region
 }
 
+# [START cloud_sql_sqlserver_instance_create_public]
 module "mssql" {
   source               = "../../modules/mssql"
```
morgante, line 24: For examples included in the docs, we should source from the registry:

```suggestion
  source  = "GoogleCloudPlatform/sql-db/google//modules/mssql"
  version = "~> 5.0"
```

tlichtenberg: changes are on the way

morgante: Thanks, please be sure to do the same for all the examples.

tlichtenberg: ok, thanks.

PR #216, opened 2021-05-11
fix: deprecates the usage of gcp-inspec and replaces integration tests with gcloud

g-awmalik:
gcp-inpsec for Cloud SQL has deprecated a set of attributes in v.1.8.0 (issue documented [here](https://github.com/inspec/inspec-gcp/issues/321)) causing integration tests to fail. This PR updates all 6 integration test to use gcloud instead for gcp-inspec.

Additionally, there's a minor changes for increasing the timeout for postgresql operations to 15m since provisioning was failing on 10m.

### DIFF
--- a/modules/postgresql/variables.tf
+++ b/modules/postgresql/variables.tf
@@ -247,19 +247,19 @@ variable "additional_users" {
 variable "create_timeout" {
   description = "The optional timout that is applied to limit long database creates."
   type        = string
-  default     = "10m"
+  default     = "15m"
 }
 
 variable "update_timeout" {
   description = "The optional timout that is applied to limit long database updates."
   type        = string
-  default     = "10m"
+  default     = "15m"
 }
 
 variable "delete_timeout" {
   description = "The optional timout that is applied to limit long database deletes."
   type        = string
-  default     = "10m"
+  default     = "15m"
 }
 
 variable "encryption_key_name" {
--- a/test/integration/mssql-ha/controls/mssql.rb
+++ b/test/integration/mssql-ha/controls/mssql.rb
@@ -12,46 +12,118 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+require 'json'
+
 project_id = attribute('project_id')
 basename   = attribute('name')
 authorized_network = attribute('authorized_network')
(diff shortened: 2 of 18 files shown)


comment-bot-dev:

modules/postgresql/variables.tf
```diff
@@ -247,19 +247,19 @@ variable "additional_users" {
 variable "create_timeout" {
   description = "The optional timout that is applied to limit long database creates."
   type        = string
-  default     = "10m"
+  default     = "15m"
```
morgante, line 250: Note we should generally be careful about changing defaults.

test/integration/mssql-ha/controls/mssql.rb
```diff
@@ -12,46 +12,118 @@
...
 basename   = attribute('name')
 authorized_network = attribute('authorized_network')
+region = "us-central1"
+
+activation_policy = "ALWAYS"
+availability_type = "REGIONAL"
+data_disk_size_gb = 10
+data_disk_type = "PD_SSD"
+kind = "sql#settings"
+pricing_plan = "PER_USE"
```
morgante, line 27: I don't think we need to test settings we're not really modifying.

morgante (approved):

modules/postgresql/variables.tf
g-awmalik: Understood. This had to be changed though because it would consistently fail on creation.

test/integration/mssql-ha/controls/mssql.rb
g-awmalik: You're right but there are few place where a subset of these are modified. I can do another PR to clean this up.

modules/postgresql/variables.tf
bharathkkb: Seems like test failures were happening on the provider too and they bumped to 30m default for the resource. I think we should align with that default https://github.com/GoogleCloudPlatform/magic-modules/pull/4636

g-awmalik: Are you suggesting we raise the timeout to 30m for all 3 modules i.e. mssql, mysql & postgresql?

Issue #201, opened 2021-03-10
IAM users/service accounts


jtatum:
Adding IAM users to cloud sql is a bit painful. Terraform has a `google_service_account` resource which gives easy access to the user's email. Unfortunately, IAM users need a stripped version of user email - per the docs:

> EMAIL: The email address of the service account. Due to length limit on database username, you need to omit the .gserviceaccount.com suffix in the email. For example, the username for the service account sa-name@project-id.iam.gserviceaccount.com should be sa-name@project-id.iam.

To get from `google_service_account.email` to `google_sql_user.name`, you need to do some fancy Terraform footwork with split, slice, join, etc. This seems like a good place for a module to help out - maybe if I give it an IAM service account, it handles that for me? Also it would be nice if the DB flags for IAM auth were enabled automatically if IAM users are supplied.

morgante:
Good suggestion, we should add that logic here: https://github.com/terraform-google-modules/terraform-google-sql-db/blob/master/modules/postgresql/main.tf#L170

g-awmalik:
I have this enhancement ready to go but blocked by [this](https://github.com/terraform-google-modules/terraform-google-sql-db/issues/211) kitchen int test issue. Seems like a gem version mismatch somewhere but haven't been able to pin point it.

bharathkkb:
fixed via #218

PR #218, opened 2021-05-12
feat: extended the postgresql module to accept IAM users and services accounts

g-awmalik:
The postgresql module has been extended to accept IAM users (both Google IDs and Service Accounts) as valid users to the database. In addition to that, this PR contains an example for provisioning IAM users and a corresponding integration test.

### DIFF
--- a/examples/postgresql-public-iam/main.tf
+++ b/examples/postgresql-public-iam/main.tf
@@ -1,0 +1,83 @@
+/**
+ * Copyright 2019 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+provider "google" {
+  version = "~> 3.22"
+}
+
+provider "google-beta" {
+  version = "~> 3.5"
+}
+
+provider "null" {
+  version = "~> 2.1"
+}
+
+provider "random" {
+  version = "~> 2.2"
+}
+
+module "postgresql-db" {
+  source               = "../../modules/postgresql"
+  name                 = var.db_name
+  random_instance_name = true
+  database_version     = "POSTGRES_9_6"
+  project_id           = var.project_id
+  zone                 = "us-central1-c"
+  region               = "us-central1"
+  tier                 = "db-f1-micro"
+
+  deletion_protection = false
+
+  ip_configuration = {
+    ipv4_enabled        = true
+    private_network     = null
+    require_ssl         = true
+    authorized_networks = var.authorized_networks
+  }
+
+  database_flags = [
+    {
+      name  = "cloudsql.iam_authentication"
+      value = "On"
+    },
+  ]
+
+  additional_users = [
+    {
+      name     = "tftest2"
+      password = "abcdefg"
+      host     = "localhost"
+    },
+    {
+      name     = "tftest3"
+      password = "abcdefg"
+      host     = "localhost"
+    },
+  ]
+
+  # Supports creation of both IAM Users and IAM Service Accounts
+  iam_users = [
+    {
+      email         = var.cloudsql_pg_sa
+      is_account_sa = true
+    },
+    {
+      email         = "dbadmin@goosecorp.org"
+      is_account_sa = false
+    },
+  ]
+}
--- a/modules/postgresql/main.tf
+++ b/modules/postgresql/main.tf
@@ -26,6 +26,7 @@ locals {
 
   databases = { for db in var.additional_databases : db.name => db }
   users     = { for u in var.additional_users : u.name => u }
+  iam_users = { for iu in var.iam_users : iu.email => iu }
 }
 
 resource "random_id" "suffix" {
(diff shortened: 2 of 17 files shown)


comment-bot-dev:

modules/postgresql/main.tf
```diff
@@ -173,6 +174,45 @@ resource "google_sql_user" "additional_users" {
...
+resource "google_service_account" "sa" {
+  for_each = {
+    for i in local.iam_users : i.name => i
+    if i.is_account_sa
+  }
+
+  project      = var.project_id
+  account_id   = each.value.name
+  display_name = each.value.description
+}
```
bharathkkb, line 203: This would create a new SA. IIUC the ask from https://github.com/terraform-google-modules/terraform-google-sql-db/issues/201 is to allow users to pass in a SA email like `sa-name@project-id.iam.gserviceaccount.com ` and have some logic within the module to convert that into ` sa-name@project-id.iam`

test/integration/postgresql-public-iam/controls/pg.rb
```diff
@@ -0,0 +1,136 @@
...
+
+  describe "postgresql_public_database" do
+    it "has 1 IAM user" do
+      expect(data.select {|k,v| k['type'] == "CLOUD_IAM_USER"}.size).to eq 1
+    end
+  end
+
+  describe "postgresql_public_database" do
+    it "has 1 IAM service account user" do
+      expect(data.select {|k,v| k['type'] == "CLOUD_IAM_SERVICE_ACCOUNT"}.size).to eq 1
```
bharathkkb, line 133: Let's also assert that our string manipulations within TF are correct, if possible, by expecting a hardcoded expected value?

bharathkkb (commented):
Thanks for the PR @g-awmalik

modules/postgresql/main.tf
g-awmalik: Yeah that's how I started implementing it but then I thought it would be much easier if the module allowed for creation of the SA as well. This way all the user has to do is pass in the sa-name and the module creates the sa (if not already present) and does the trimming logic necessary to add it to postgresql as a user.

test/integration/postgresql-public-iam/controls/pg.rb
g-awmalik: Sure, I can add that check as well.

modules/postgresql/main.tf
g-awmalik: And I believe they are using the `google_service_account` resource in their module to provision an SA. I think the whole creation & trimming of the SA should be abstracted from the user and done by the module as long as they pass in a sa-name.

morgante: We should not create service accounts in this module.

The general principle is that service accounts should be created where they are *used* (ie. near the resource which will use it as an "identity") not where they will be granted permission.

g-awmalik: That makes sense to me now. I'll change this accordingly.

g-awmalik:
@bharathkkb - please take a look at the changes now. Thnx.

examples/postgresql-public-iam/main.tf
```diff
@@ -0,0 +1,83 @@
...
+      password = "abcdefg"
+      host     = "localhost"
+    },
+  ]
+
+  # Supports creation of both IAM Users and IAM Service Accounts
+  iam_users = [
+    {
+      email         = var.cloudsql_pg_sa
+      is_account_sa = true
```
bharathkkb, line 76: Would it be possible to do this automatically? SA would be of the form `...gserviceaccount.com` so perhaps `is_sa = trimsuffix(var. email"gserviceaccount.com")==var. email`

modules/postgresql/main.tf
```diff
@@ -173,6 +174,33 @@ resource "google_sql_user" "additional_users" {
...
+resource "google_project_iam_member" "iam_binding" {
+  for_each = local.iam_users
+  project  = var.project_id
+  role     = "roles/cloudsql.instanceUser"
+  member = each.value.is_account_sa ? (
+    "serviceAccount:${each.value.email}"
+    ) : (
+    "user:${each.value.email}"
+  )
+}
```
bharathkkb, line 202: nit: suggest we move this up a block for readability as the previous block depends on this `google_project_iam_member.iam_binding,`

bharathkkb (commented):
Overall LGTM

/cc @morgante for another look

examples/postgresql-public-iam/main.tf
g-awmalik: That's actually being done in the model [here](https://github.com/terraform-google-modules/terraform-google-sql-db/pull/218/files#diff-5af99345140d405f1b9683608683cd8e45b6cce368ba0446706651ca3a92f041R181). I wanted to keep the input as simple as possible and let the module do the trimming/formatting etc. Let me know what you think.

modules/postgresql/main.tf
g-awmalik: sure

examples/postgresql-public-iam/main.tf
g-awmalik: Rethinking this and I think you have a valid point. If the email is an SA or not can be calculated through logic and I'd rather put that logic in the module as well and let the user just provide the email. That way we can get rid of that that attribute as well. I'll make that change.

bharathkkb: @g-awmalik I was thinking if we can detect if its an SA or not without needing user input.

My idea was if `trimsuffix(each.value.email, ".gserviceaccount.com") != each.value.email` that would mean that `trimsuffix` was able to trim `.gserviceaccount.com` which means it is a service account. If both are equal that means `trimsuffix` failed to trim because no  ".gserviceaccount.com" suffix and hence user email. WDYT?

```tf
name =  trimsuffix(each.value.email, "gserviceaccount.com") != each.value.email ? ( trimsuffix(each.value.email, ".gserviceaccount.com")    ) : (    each.value.email  )
```

g-awmalik: Yeah totally agree. I think we crossed each other in our latest responses. I get it and I'll make that change. The user will only provide an email with that.

modules/postgresql/main.tf
g-awmalik: done

examples/postgresql-public-iam/main.tf
g-awmalik: done.

Issue #222, opened 2021-05-26
Postgres Module: additional_users requires password


Ulminator:
When trying to create users with the `additional_users` variable it looks like the desired functionality is for the user's passwords to be defaulted to `random_id.user-password.hex`, but this is not working for me. When I don't provide a password for a user in that list of maps, I receive an invalid value error for the variable. Since I now have to provide a password key to the user maps, the following lookup never reaches the "default" option.

```terraform
resource "google_sql_user" "additional_users" {
  for_each  = local.users
  project     = var.project_id
  name        = each.value.name
  password = lookup(each.value, "password", random_id.user-password.hex)
  instance   = google_sql_database_instance.default.name
  depends_on = [null_resource.module_depends_on, google_sql_database_instance.default]
}
```

morgante:
Have you tried setting the value to `null`? Like this:

```
additional_users = [{name: "test", password: null}]
```

Ulminator:
Yeah when I tried that earlier it failed as well with this error message:

```
Error: Error, failed to insert user test_user into instance test_instance: googleapi: Error 400: Invalid request: Missing user password for PostgreSQL instance., invalid

  on main.tf line 171, in resource "google_sql_user" "additional_users":
 171: resource "google_sql_user" "additional_users" {
  ```

morgante:
Interesting, this is definitely a bug then. `null` should be sufficient to accept the default.

We'll put it on the queue, but would also be happy to review a PR.

imrannayer:
Fixed in #398

PR #223, opened 2021-05-27
Postgres additional_users: prevent null from being passed as password

Ulminator:
This is a fix for the issue referenced [here](https://github.com/terraform-google-modules/terraform-google-sql-db/issues/222). The cause was due to terraform interpreting `null` as a literal value instead of using the default value. It seems to have been an issue for awhile and is referenced in this [terraform thread](https://github.com/hashicorp/terraform/issues/21702).


comment-bot-dev:

morgante (approved):

PR #251, opened 2021-10-07
feat: Add CMEK support for cross-region read replicas

### LINKED ISSUE #249: PostgresSQL CMEK issue with cross region read replica
While creating cloud SQL instance with cross region read replica module throw error for creating replica instance. reason that for read replica there is no option to define CMEK (for a different region). 

following is snippet for error -
Error: Error, failed to create instance MY-INSTANCE-REPLICA-TEST0: googleapi: Error 400: Invalid request: Disk encryption config is not specified for the cross region replica.., invalid
│
│   with module.tenant_postgressql.google_sql_database_instance.replicas["my-i3-replica-test0"],
│   on ./../../../modules/postgresql/read_replica.tf line 23, in resource "google_sql_database_instance" "replicas":
│   23: resource "google_sql_database_instance" "replicas" {

imrannayer:
Fix #249 
It is a breaking change. variable `read_replicas` has an extra key `encryption_key_name`


comment-bot-dev:

jmymy:
Anyone experiencing weird issues with this?

I have been passing in the var, and now its saying the read replica KMS key will be set to `null` and requires replica replacement. I let it destroy, and give it another name or wait 7 days for the name to come available, and It creates the new replica instance just fine. However, on the next run, it goes to destroy it again.

jmymy:
I had replicas created before this version bump, they were already getting the `encryption_key_name` passed in as a value, so the state file has the `encryption_key_name` set. With this bump, its now trying to set that value to null, resulting in a replica recreation.

jmymy:
`encryption_key_name  = (join("-", slice(split("-", lookup(each.value, "zone", var.zone)), 0, 2))) == var.region ? null : each.value.encryption_key_name`
https://github.com/terraform-google-modules/terraform-google-sql-db/pull/251/files#diff-2961a2500544e56735fcf3af9d494ba4c33d54dba1211b88782b0c1c4805d81cR32

appears to be the culprit. There is no way i can continue without re-creating the replicas and ensuring that I pass in `null` the next creation. If i dont pass in null for replicas, the instance still gets created as the provider/api still accept it and the following TF runs will recreate it and be in a loop.

morgante:
@jmymy Can you open an issue w/ the details? Would also be happy to review a PR.

Issue #264, opened 2021-12-14
Cloud SQL (mysql and postgres) modules timeouts on user creation with replicas


romankaminski:
### TL;DR

Creating postgres database with replica with v8.0.0 results in timeout creating sql user ( on "google_sql_user"). 
There is default timeout for google_sql_user = 10minutes. Cloning your module and expanding the timeout to 20m solved the problem.

### Expected behavior

"google_sql_user" should have timeouts parametrized, so we can expand the time needed for the creation.

### Observed behavior

timeout creating user for replica, what is failing the whole IaC

### Terraform Configuration

```hcl
testdata

project_id        = "******************"
instance_name     = "pgsql-testcase"
region            = "us-central1"
zone              = "us-central1-c"
availability_type = "REGIONAL"
user_labels = {
  environment_label = "pgsql_el_test"
  atosmanaged       = "true"
}
db_version          = "POSTGRES_13"
deletion_protection = "false"
backup_configuration = {
  enabled                        = "true"
  start_time                     = "3:30"
  location                       = "us-central1"
  point_in_time_recovery_enabled = "true"
  transaction_log_retention_days = "7"
  retained_backups               = null
  retention_unit                 = null
}
ip_configuration = {
  authorized_networks = []
  ipv4_enabled        = "false"
  require_ssl         = "true"
  private_network     = "projects/prj-d-shared-base-430c/global/networks/vpc-d-shared-base"
}
read_replicas = [{
    name = "-0"
    zone = "us-west1-a"
    tier = "db-f1-micro"
    disk_type = "PD_HDD"
    disk_autoresize = "false"
    disk_size = "10"
    user_labels = { environment = "test" }
    database_flags = [{
            name = "log_temp_files", value = 0
        },
        {
            name = "log_checkpoints", value = "on"
        },
        {
            name = "log_connections", value = "on"
        },
        {
            name = "log_disconnections", value = "on"
        },
       {
            name = "log_min_duration_statement", value = "-1"
        }]
    ip_configuration = {
        authorized_networks = []
        ipv4_enabled    = "false"
        private_network = "projects/************/global/networks/vpc-d-shared-base"
        require_ssl     = "true"
    }
    encryption_key_name = ""
}]
```

### Terraform Version

```sh
Terraform v1.0.10
on linux_amd64
```

### Additional information

It seems the problem didn't appear on v7.1.0 we had before and I see "provider" has been added to postgres database, so could this be effect of the google-beta provider?

[...]

PR #268, opened 2021-12-29
fix: add depends on replicas for user creation

g-awmalik:
This is an interim fix for [this issue](https://github.com/terraform-google-modules/terraform-google-sql-db/issues/264)
Additionally, it uses the custom machine types for the postgresql examples since the shared instance was taking too long and timing out.

### DIFF
--- a/examples/postgresql-public-iam/main.tf
+++ b/examples/postgresql-public-iam/main.tf
@@ -23,7 +23,7 @@ module "postgresql-db" {
   project_id           = var.project_id
   zone                 = "us-central1-c"
   region               = "us-central1"
-  tier                 = "db-f1-micro"
+  tier                 = "db-custom-2-13312"
 
   deletion_protection = false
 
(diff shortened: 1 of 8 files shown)


comment-bot-dev:

morgante (changes requested):
Please do not use timeouts for this. That's not the right fix.

If there are issues with parallelism, the correct fix is to adjust the `depends_on` so it doesn't start until the earlier resources are done.

We should *never* expose timeout parameters to users.

bharathkkb (changes requested):
We should do this for [mysql](https://github.com/terraform-google-modules/terraform-google-sql-db/tree/master/modules/mssql) too

g-awmalik:
@bharathkkb - I can do that as another PR. Let's let this go through to unblock the customer for now.

bharathkkb (approved):

morgante:
@g-awmalik Can you add this for mysql and mssql? I want to get those into the release.

g-awmalik:
@morgante - this includes the change for MySql as well. However, this module doesn't include replica setup for MSSQL yet. I can work on that but we may want to push these changes for now.

PR #294, opened 2022-04-19
Add sensitive flag for two outputs

hjoh95:
Running a terraform plan using the postgresql module resulted in two errors for these outputs. I believe this is the expected fix

### DIFF
--- a/modules/postgresql/outputs.tf
+++ b/modules/postgresql/outputs.tf
@@ -112,6 +112,7 @@ output "additional_users" {
 output "primary" {
   value       = google_sql_database_instance.default
   description = "The `google_sql_database_instance` resource representing the primary instance"
+  sensitive   = true
 }
 
 output "replicas" {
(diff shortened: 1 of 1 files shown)


comment-bot-dev:
@hjoh95

morgante:
This is technically not required, because this module is not meant as a root module, but we can add it anyways.

morgante (approved):

PhilBrammer:
Isn't this a breaking change, @morgante?  I'm assuming that there are implementations that automate/screen against TF APPLY output.

morgante:
@PhilBrammer I wouldn't consider a sensitive flag a breaking change. It shouldn't impact any automated configuration to switch an output to sensitive, since Terraform can still reference sensitive outputs.

PhilBrammer:
Not directly.  Working with @hjoh95 on this, but seems to me if I were using this module, I would've been getting a list of instances created as output.  Now those would be omitted, correct?  And to me, that seems like a breaking change for any consumer expecting this output.  Sure they can adapt, but am wondering if this approach is safe "just pushing through" when these modules were being used in the wrong way anyhow?  @hjoh95 created a wrapper to use this module correctly.

PR #297, opened 2022-04-26
use random_password instead of random_id

mikecook:
- random_id byte_length = 8 (integers) contains 26.6 bits of
  entropy. hex encoding does not change that entropy.
  Instead use random_password length = 32 restricted to (upper,
  lower, int) which contains 190.5 bits of entropy.
- Restrict random_password to special = false to prevent issues
  with allowed characters.

### DIFF
--- a/modules/mysql/main.tf
+++ b/modules/mysql/main.tf
@@ -152,23 +152,25 @@ resource "google_sql_database" "additional_databases" {
   depends_on = [null_resource.module_depends_on, google_sql_database_instance.default]
 }
 
-resource "random_id" "user-password" {
+resource "random_password" "user-password" {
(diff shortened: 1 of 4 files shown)


bharathkkb (commented):
Thanks for the PR  @mikecook 
Lets also add an upgrade guide https://github.com/terraform-google-modules/terraform-google-sql-db/tree/master/docs

mikecook:
I don't use this module and don't have a place to test it. I don't think I'm the correct person to write an upgrade guide.

bharathkkb:
fixed via https://github.com/terraform-google-modules/terraform-google-sql-db/pull/308

PR #301, opened 2022-05-09
added module_depends_on to private_service_access

### LINKED ISSUE #90: Add module_depends_on resource to private_service_access
- `private_service_access` module depends on *Service Networking API* and `vpc_network`, so it would be very helpful to have **module_depends_on** in `private_service_access`. 
- Also description for `private_service_access` should be updated, it useful not only for MySQL but also for Postgres databases.

imrannayer:
fix #90


comment-bot-dev:
@imrannayer

imrannayer:
Make sense. I think we can close this PR as they can use native support directly in root module.

PR #313, opened 2022-06-10
Update random provider to 3.2

midu-git:


comment-bot-dev:
@midu-git

midu-git:
random provider has introduced [bcrypt_hash](https://registry.terraform.io/providers/hashicorp/random/3.2.0/docs/resources/password) as output to random_password in version 3.2. We use this output in our setup making this module not usable for us. However, there is no requirement in this module to stick on version 3.1.

bharathkkb:
@midu-git IIUC you would like to v3.2 of the provider. This module should support any version greater than 3.1, so we would not need to bump the min verson.

Issue #319, opened 2022-07-10
Output refers to sensitive values


6aKa:
### TL;DR

```
│ Error: Output refers to sensitive values
│
│   on outputs.tf line 118:
│  118: output "replicas" {
│
│ To reduce the risk of accidentally exporting sensitive data that was
│ intended to be only internal, Terraform requires that any root module
│ output containing sensitive data be explicitly marked as sensitive, to
│ confirm your intent.
│
│ If you do intend to export this data, annotate the output value as
│ sensitive by adding the following argument:
│     sensitive = true
```

### Terraform Configuration

```hcl
default
```

### Terraform Version

```sh
Terraform v1.2.4
```

### Additional information

postgres module version 11.0.0

6aKa:
fixed

PR #334, opened 2022-08-05
fix output refers to sensitive values

6aKa:
Fix for issue https://github.com/terraform-google-modules/terraform-google-sql-db/issues/319


comment-bot-dev:
@6aKa

bharathkkb (commented):
Thanks for the PR @6aKa. Please sign the CLA.
Also it is worth noting that this module is not meant to be directly used as the root module and it is only in root modules we need to mark outputs as sensitive. We can still accept this since we have this for the mysql.

6aKa:
I catch error when try use module from terragrunt.
```
terraform {
  source = "tfr:///GoogleCloudPlatform/sql-db/google//modules/postgresql//.?version=11.0.0"
}
inputs = {
  name                               = "test-db"
  random_instance_name = true
  ...
}
```
Because terragrunt catch all output from terraform module.

comment-bot-dev:
@6aKa

bharathkkb (approved):

PR #355, opened 2022-10-20
Db master and replica update solution for MYSQL

NA2047:
I have observed an issue where updated the DB to a later version with read replicas enabled causes an error or deletes the replicates and the master depending on the version. The solution I'am purposing allows the user to update in stages first the replicas then the master


comment-bot-dev:
@NA2047
Thanks for the PR! 🚀
- ⚠️check_documentation
 The documentation needs to be regenerated. Please run `make generate_docs`. 
 ```bash 
Checking for documentation generation
diff -r '--exclude=.terraform' '--exclude=.kitchen' '--exclude=.git' '--exclude=autogen' '--exclude=*.tfvars' /workspace/modules/mysql/README.md /tmp/tmp.RvvJrjGWlI/generate_docs/workspace/modules/mysql/README.md
17a18
> | database\_version\_replica | The read replica database version to use. There is no need to set this, the master database version will be taken by default. However to update databases, read replicas must be updated first, setting this to an update version will cause the replica to update, then you may update the master with the var database\_version and leave or remove this field after | `string` | `null` | no |
Error: Documentation generation has not been run, please run the
'make docker_generate_docs' command and commit the above changes.
 ``` 
- ⚠️check_terraform
 ```bash 
Running terraform fmt
modules/mysql/variables.tf
--- old/modules/mysql/variables.tf
+++ new/modules/mysql/variables.tf
@@ -33,7 +33,7 @@
 variable "database_version_replica" {
   description = "The read replica database version to use. There is no need to set this, the master database version will be taken by default. However to update databases, read replicas must be updated first, setting this to an update version will cause the replica to update, then you may update the master with the var database_version and leave or remove this field after"
   type        = string
-  default = null
+  default     = null
 }
 
 // required
 ```

comment-bot-dev:
@NA2047

modules/mysql/variables.tf
```diff
@@ -30,6 +30,12 @@ variable "random_instance_name" {
   default     = false
 }
 
+variable "database_version_replica" {
+  description = "The read replica database version to use. There is no need to set this, the master database version will be taken by default. However to update databases, read replicas must be updated first, setting this to an update version will cause the replica to update, then you may update the master with the var database_version and leave or remove this field after"
+  type        = string
+  default     = null
```
g-awmalik, line 36: For strings, it's better to have the default as: ""

modules/mysql/variables.tf
```diff
@@ -30,6 +30,12 @@ variable "random_instance_name" {
   default     = false
 }
 
+variable "database_version_replica" {
+  description = "The read replica database version to use. There is no need to set this, the master database version will be taken by default. However to update databases, read replicas must be updated first, setting this to an update version will cause the replica to update, then you may update the master with the var database_version and leave or remove this field after"
```
g-awmalik, line 34: You can reduce the description here. E.g. the fact that you a `default` for this implies that this is optional.

g-awmalik (commented):
Thanks for the PR. Can you add this functionality to an example as well to make sure it's working end-to-end?

You could probably add it [here](https://github.com/terraform-google-modules/terraform-google-sql-db/blob/master/examples/mysql-ha/main.tf#L83) by adding `database_version` to one of the replicas and leaving the other two as is. That'll confirm the logic.

modules/mysql/variables.tf
NA2047: Sounds good!

comment-bot-dev:
@NA2047

comment-bot-dev:
@NA2047

modules/mysql/variables.tf
```diff
@@ -30,6 +30,12 @@ variable "random_instance_name" {
   default     = false
 }
 
+variable "database_version_replica" {
```
g-awmalik, line 33: This variable should be part of the [`read_replicas`](https://github.com/terraform-google-modules/terraform-google-sql-db/blob/fa0d938a582febe6e4c3ce4cdc6aaa0b7a3b221e/modules/mysql/variables.tf#L181) slice since each replica can be upgraded individually

modules/mysql/read_replica.tf
```diff
@@ -25,7 +25,7 @@ resource "google_sql_database_instance" "replicas" {
   for_each             = local.replicas
   project              = var.project_id
   name                 = "${local.master_instance_name}-replica${var.read_replica_name_suffix}${each.value.name}"
-  database_version     = var.database_version
+  database_version     = var.database_version_replica != "" ? var.database_version_replica : var.database_version
```
g-awmalik, line 28: This logic should look at the `read_replicas` slice and try to pull the `database_version` from it. If not found, the `database_version` for the primary should be used.

examples/mysql-ha/main.tf
```diff
@@ -80,6 +80,7 @@ module "mysql" {
 
   // Read replica configurations
   read_replica_name_suffix = "-test"
+  database_version_replica = "MYSQL_5_7"
```
g-awmalik, line 83: This should be set within the `read_replicas` slice.

g-awmalik, line 83: nit: let's call this replica_database_version

g-awmalik (commented):
I realized we're using a strongly typed object for replicas vs. a map which makes making certain attributes optional difficult. Let's keep this approach that you proposed for now and make the int tests work. I'll approve it as soon as they all pass. Thanks again!

comment-bot-dev:
@NA2047
Thanks for the PR! 🚀
- ⚠️check_documentation
 The documentation needs to be regenerated. Please run `make generate_docs`. 
 ```bash 
Checking for documentation generation
diff -r '--exclude=.terraform' '--exclude=.kitchen' '--exclude=.git' '--exclude=autogen' '--exclude=*.tfvars' /workspace/modules/mysql/README.md /tmp/tmp.45Gm2TklVQ/generate_docs/workspace/modules/mysql/README.md
18d17
< | database\_version\_replica | The read replica database version to use. This var should only be used during a database update. The update sequence 1. read-replica 2. master, setting this to an updated version will cause the replica to update, then you may update the master with the var database\_version and remove this field after update is complete | `string` | `""` | no |
43a43
> | replica\_database\_version | The read replica database version to use. This var should only be used during a database update. The update sequence 1. read-replica 2. master, setting this to an updated version will cause the replica to update, then you may update the master with the var database\_version and remove this field after update is complete | `string` | `""` | no |
Error: Documentation generation has not been run, please run the
'make docker_generate_docs' command and commit the above changes.
 ```

comment-bot-dev:
@NA2047

NA2047:
> I realized we're using a strongly typed object for replicas vs. a map which makes making certain attributes optional difficult. Let's keep this approach that you proposed for now and make the int tests work. I'll approve it as soon as they all pass. Thanks again!

I believe they are failing due to the 10min Timeout duration for destroying the resources for the test. I encountered a simular issue during my testing. I believe this has to be handled on the provider however on the async call

comment-bot-dev:
@NA2047

g-awmalik (approved):
LGTM

Issue #373, opened 2022-11-10
PostgresSQL 14 DBs are requesting a  root password 


kev-pinto-cts:
### TL;DR

The Module works fine with Postgres database Version 10, however when one uses Version 14(>=11), it throws an error stating 
Error: Error, failed to create instance pginst: googleapi: Error 400: Missing parameter: Root Password., required

The TF page has the following note which seems related to the issue
NOTE on google_sql_database_instance:
- Second-generation instances include a default 'root'@'%' user with no password. This user will be deleted by Terraform on instance creation. You should use google_sql_user to define a custom user with a restricted host and strong password.

### Expected behavior

It should have created the Instance with PostgresSQL 14

### Observed behavior

module.sql-db_postgresql["pginst"].google_sql_database_instance.default: Creating...
╷
│ Error: Error, failed to create instance pginst: googleapi: Error 400: Missing parameter: Root Password., required
│ 
│   with module.sql-db_postgresql["pginst"].google_sql_database_instance.default,
│   on .terraform/modules/sql-db_postgresql/modules/postgresql/main.tf line 48, in resource "google_sql_database_instance" "default":
│   48: resource "google_sql_database_instance" "default" {

### Terraform Configuration

```hcl
# backend.tf
terraform {
  backend "gcs" {
    bucket = "XXX-99999-tfstate"
    prefix = "cloudsql"
  }
}

#provider.tf
terraform {
  required_providers {
    google = {
      source  = "hashicorp/google"
      version = "4.28.0"
    }
  }
}

provider "google" {
  project = "XXX"
  region  = "europe-west2"
}

#variables.tf
variable "location" {}

#terraform.tfvars
location = "europe-west2"

#main.tf
locals {
  config                   = yamldecode(file("./config/postgres.yaml"))
  default_user_name        = "postgres"
  default_database_version = "14"
  default_user_labels      = {
    project = data.google_project.nwmmr.project_id
    region  = var.location
  }
}

data "google_project" "nwmmr" {
}

module "sql-db_postgresql" {
  source   = "GoogleCloudPlatform/sql-db/google//modules/postgresql"
  version  = "13.0.0"
  for_each = {for instance in local.config.instances : instance.name => instance}

  name                 = each.value.name
  project_id           = data.google_project.nwmmr.project_id
  region               = try(each.value.region, var.location)
  zone                 = "${var.location}-a"
  deletion_protection  = false
[...]

PR #388, opened 2022-12-12
fix: Added functionality to simply specify the database version number ins…

ravisiddhu:
This PR is a fix for https://github.com/terraform-google-modules/terraform-google-sql-db/issues/373

Buganizer : b/259913441

The reason for the issue is that when using postgres module, customer assumes that since they have explicitly chosen postgres module, they only need to specify the database version number instead of the whole version name as it says in doc, but that's not the case. (Example "14" instead of "POSTGRES_14")

When customer sets the  database version field as simply "14" instead of "POSTGRES_14", the instances.insert API is actually trying to create a SQL Server instance (by default the database version used is "SQLSERVER_2017_ENTERPRISE"), that is why the API is throwing an error saying "Root password is required", because SQL Server instances requires root password.

But the customer assumptions seems valid and reasonable, hence I have added a functionality in postgres module which would prepend "POSTGRES_" to database version if they have only specified the numeric part of the database version.
Example "14" to "POSTGRES_14"

Successfully passed docker_test_lint


comment-bot-dev:
@ravisiddhu

modules/postgresql/main.tf
```diff
@@ -49,7 +49,7 @@ resource "google_sql_database_instance" "default" {
   provider            = google-beta
   project             = var.project_id
   name                = local.master_instance_name
-  database_version    = var.database_version
+  database_version    = (length(var.database_version) >= 8 && (substr(var.database_version, 0, 8) == "POSTGRES" || substr(var.database_version, 0, 8) == "postgres")) ? var.database_version : format("POSTGRES_%s", var.database_version)
```
bharathkkb, line 52: IIUC the logic is if it is number, prepend  `POSTGRES_` else pass var.database_version. Could we simplify the logic to something like `can(tonumber(var. database_version)) ? "POSTGRES_${var. database_version}" : var. database_version `. 

We could also add [variable validation](https://developer.hashicorp.com/terraform/language/values/variables#custom-validation-rules) rules to `var.database_version` to enforce one of number or string that starts with `POSTGRES_`.

bharathkkb (commented):
Thanks for the PR @ravisiddhu

comment-bot-dev:
@ravisiddhu

modules/postgresql/main.tf
ravisiddhu: I will add variable validation rules, but "tonumber" function would not work if database version is 9_6 though it is a valid representation of database version, right?

comment-bot-dev:
@ravisiddhu

modules/postgresql/main.tf
ravisiddhu: I have added variable validation rules for "database_version", which enforces that the version is of the pattern "POSTGRES_9_6" or "postgres_14" or "9_6".

After the validation, the main.tf file will check whether the database_version starts with a number, then it prepends it with "POSTGRES_" or returns the database version as it is.

modules/postgresql/variables.tf
```diff
@@ -34,6 +34,11 @@ variable "random_instance_name" {
 variable "database_version" {
   description = "The database version to use"
   type        = string
+
+  validation {
+    condition     = (length(var.database_version) >= 9 && ((substr(var.database_version, 0, 9) == "POSTGRES_" || substr(var.database_version, 0, 9) == "postgres_") && can(regex("^\\d+(?:_?\\d)*$", substr(var.database_version, 9, -1))))) || can(regex("^\\d+(?:_?\\d)*$", var.database_version))
```
bharathkkb, line 39: nit: I think we can just lower case to keep this shorter
```suggestion
    condition     = (length(var.database_version) >= 9 && (lower(substr(var.database_version, 0, 9) && can(regex("^\\d+(?:_?\\d)*$", substr(var.database_version, 9, -1))))) || can(regex("^\\d+(?:_?\\d)*$", var.database_version))
```

modules/postgresql/main.tf
bharathkkb: @ravisiddhu good point, we could do a replace for `_` but your new approach looks good.

comment-bot-dev:
@ravisiddhu
Thanks for the PR! 🚀
- ⚠️check_terraform
 ```bash 
Running terraform fmt
modules/postgresql/main.tf
--- old/modules/postgresql/main.tf
+++ new/modules/postgresql/main.tf
@@ -49,7 +49,7 @@
   provider            = google-beta
   project             = var.project_id
   name                = local.master_instance_name
-  database_version    = can(regex("\d", substr(var.database_version, 0, 1))) ? format("POSTGRES_%s", var.database_version) : replace(var.database_version, substr(var.database_version, 0,8), "POSTGRES")
+  database_version    = can(regex("\d", substr(var.database_version, 0, 1))) ? format("POSTGRES_%s", var.database_version) : replace(var.database_version, substr(var.database_version, 0, 8), "POSTGRES")
   region              = var.region
   encryption_key_name = var.encryption_key_name
   deletion_protection = var.deletion_protection
 ```

comment-bot-dev:
@ravisiddhu
Thanks for the PR! 🚀
- ⚠️check_terraform
 ```bash 
Running terraform fmt
modules/postgresql/main.tf
--- old/modules/postgresql/main.tf
+++ new/modules/postgresql/main.tf
@@ -49,7 +49,7 @@
   provider            = google-beta
   project             = var.project_id
   name                = local.master_instance_name
-  database_version    = can(regex("\d", substr(var.database_version, 0, 1))) ? format("POSTGRES_%s", var.database_version) : replace(var.database_version, substr(var.database_version, 0,8), "POSTGRES")
+  database_version    = can(regex("\d", substr(var.database_version, 0, 1))) ? format("POSTGRES_%s", var.database_version) : replace(var.database_version, substr(var.database_version, 0, 8), "POSTGRES")
   region              = var.region
   encryption_key_name = var.encryption_key_name
   deletion_protection = var.deletion_protection
 ```

modules/postgresql/variables.tf
ravisiddhu: I have made the changes based on your suggestion.

I have also added "replace(var.database_version, substr(var.database_version, 0,8), "POSTGRES")" instead of just "var.database_version", to avoid terraform drift. (This drift was because "postgres_14" was being converted to "POSTGRES_14" by the API, hence there was a difference in state file and config file, hence the fix.)

comment-bot-dev:
@ravisiddhu

comment-bot-dev:
@ravisiddhu

comment-bot-dev:
@ravisiddhu

comment-bot-dev:
@ravisiddhu

bharathkkb (approved):

PR #339, opened 2022-08-24
feat!: Add `secondary_zone` to db modules

jamiezieziula:
Closes #320

Issue #326, opened 2022-07-28
Add support for settings.location_preference.secondary_zone and settings.location_preference.follow_gae_application


philip-harvey:
### TL;DR

Provider v4.28.0 adds support for settings.location_preference.secondary_zone

follow_gae_application has been in the provider since at least version 3.0.0 

This module does not currently support setting these.

KrustyHack:
Ay,

It would be nice, as right now Terraform is always detecting drift when I apply my stack :

```
Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
  ~ update in-place

Terraform will perform the following actions:

  # module.postgresql_bits.module.postgresql.google_sql_database_instance.default will be updated in-place
  ~ resource "google_sql_database_instance" "default" {
        id                            = "xxx"
        name                          = "xxx"
        # (12 unchanged attributes hidden)

      ~ settings {
            # (10 unchanged attributes hidden)

          ~ location_preference {
              - secondary_zone = "europe-west1-d" -> null
                # (1 unchanged attribute hidden)
            }

            # (6 unchanged blocks hidden)
        }

        # (1 unchanged block hidden)
    }

Plan: 0 to add, 1 to change, 0 to destroy.
```

g-awmalik:
This is partially address by #339. For `settings.location_preference.follow_gae_application` if you could raise a PR, that'd be great.

philip-harvey:
remove stale

PR #390, opened 2022-12-14
feat: add `secondary_zone`, `follow_gae_application` to safer_mysql

alexrjones:
Partially addresses https://github.com/terraform-google-modules/terraform-google-sql-db/issues/326.

`secondary_zone` was added in TPG [4.28.0](https://github.com/hashicorp/terraform-provider-google/blob/main/CHANGELOG.md#4280-july-11-2022), as mentioned [here](https://github.com/terraform-google-modules/terraform-google-sql-db/pull/339/files#r989316573), hence the provider bump.


bharathkkb (approved):
Thanks for the PR @alexrjones 
Although we are bumping the providers, not marking this as a breaking release as the resolved provider version was indirectly 4.28+ due to the `mysql` module bumped in the last major version.

PR #329, opened 2022-08-02
feat!: adds availability_type for read_replicas

g-awmalik:
Addresses issue #328

PR #288, opened 2022-03-24
feat!: Add support for setting disk_autoresize_limit

kostiantyn-nemchenko:
This PR allows configuring the `disk_autoresize_limit` property available in the [google_sql_database_instance](https://github.com/hashicorp/terraform-provider-google/blob/main/google/resource_sql_database_instance.go#L257-L262) resource since Google provider version [3.70.0](https://github.com/hashicorp/terraform-provider-google/blob/main/CHANGELOG.md#3700-june-01-2021).

Adding this property to Cloud SQL modules will allow users to set a specific limit on how large the storage on their instance can automatically grow.

PR #277, opened 2022-02-10
Add allocated_ip_range property to CloudSQL modules

srs2210:
### TL;DR

The `allocated_ip_range` property was added to [google_sql_database_instance](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/sql_database_instance#allocated_ip_range) resource from Google provider version `4.4.0`. Adding this property to Cloud SQL modules will allow users to select a specific allocated range for their private instances.

### Issues fixed

- Fixes terraform-google-modules/terraform-google-sql-db#267

PR #398, opened 2022-12-22
feat!: Aligned the behaviour of additional_users resource in all 3 Cloud SQL instance modules.

ravisiddhu:
This PR is in response to https://github.com/terraform-google-modules/terraform-google-sql-db/issues/222

Buganizer bug : http://b/259914601

In mssql module, there was no way to generate random password for the user. the password was a required field, hence couldn't omit it.
If password is set as null, then no password is created for the user, and if the password is set as "", then the password is also literally "" (empty password). 

I am aligning the behaviour of `additonal_users` resource in all 3 cloud sql modules.

This PR include **BREAKING CHANGES** :- 
1) Addition of new field called `random_password` in `additional_users` resource.
2) Added support for creation of IAM users in MySQL.
2) The same modified `additional_users` resource is used in all the 3 modules.

Please merge this PR with breaking changes only when we are releasing the next major version.


comment-bot-dev:
@ravisiddhu

comment-bot-dev:
@ravisiddhu

comment-bot-dev:
@ravisiddhu

comment-bot-dev:
@ravisiddhu

comment-bot-dev:
@ravisiddhu

comment-bot-dev:
@ravisiddhu

modules/mssql/variables.tf
```diff
@@ -237,11 +237,8 @@ variable "user_password" {
 
 variable "additional_users" {
   description = "A list of users to be created in your cluster"
-  type = list(object({
-    name     = string
-    password = string
-  }))
-  default = []
+  type        = list(map(any))
+  default     = []
```
bharathkkb, line 241: We had received some feedback that not having vars typed hurt UX as users were unable to determine what values to pass in. WDYT about switching to use [optionals](https://developer.hashicorp.com/terraform/language/expressions/type-constraints#optional-object-type-attributes) instead? This will be breaking change requiring 1.3.0 but we expect most modules to switch to use optionals next year.

ravisiddhu: I don't think we should force all our customers to upgrade terraform to 1.3.0+, for a small fearure fix. I think we should wait and only cause this breaking change when it is absolutely necessary, wdyt?

As an alternate solution we can have one more resource called "additional_users_with_random_password", where users can set null or "" to the password field to generate random password (replicating the behavior of postgres module) in mssql module. In this way we are not breaking anything for the exisiting mssql customers, what do you think? Should I go forward with this?

bharathkkb: We will likely switch to 1.3 in the next breaking release as we had multiple breaking major releases ([12.0](https://github.com/terraform-google-modules/terraform-google-sql-db/pull/329),[11.0](https://github.com/terraform-google-modules/terraform-google-sql-db/pull/288),[10.0](https://github.com/terraform-google-modules/terraform-google-sql-db/pull/277),[8.0](https://github.com/terraform-google-modules/terraform-google-sql-db/pull/251)) due to changes in object types which 1.3 will avoid. Agreed we don't need to force the switch for this feature, I was thinking of bundling this with next major if we decided to use optionals.

With the alternate approach using `additional_users_with_random_password`, would we need a new input to target this? Also re reading the description, I suspect most users may not want to use a literal `""` since no password is supported by `null`. It might just be cleaner to not support literal `""`.

ravisiddhu: Setting password as "", creates an user with an empty password, then that user can login without providing password.

1) Should I use Optionals and send the PR, where you can merge this PR once we decide to create a breaking release. or
2)  With additional_users_with_random_password, yes, it would be a new input field for the customer. Now mssql customers could either use additonal_users or additonal_users_with_random_password input fields for creating users. The mssql customers can use the new input if they require their users to be set with an random password (and they still would have the option to change the random password if needed. This new field would replicate the behaviour of additional_users field in postgres module).

Should I go ahead with (1) or (2) ?

bharathkkb: > If password is set as null, then no password is created for the user, and if the password is set as "", then the password is also literally "" (empty password).
> >Setting password as "", creates an user with an empty password, then that user can login without providing password. 

In that case, what is the difference between empty string `""` as password vs no password at all via `null`? 

I am wondering even with optionals we will need to provide a default value signifying when to generate the random password - would we use empty string or null?

ravisiddhu: In postgres module, random password is generated during empty string and null values for password, but for MySQL, random password is generated only when password field is not set in the additional_users resource.
```
additional_users([{name: "test-user"}])
```
I think we can also use optionals in mssql to make the password variable optional, so that the mssql users would also have an option for not setting the password, so that a random password is only generated when the password field is not set. If we generate random password for a null or an empty string password for a user, then it would be a breaking change, since some mssql customers would be setting the password as null or "" to create  a login user without pwd.

comment-bot-dev:
@ravisiddhu

comment-bot-dev:
@ravisiddhu

comment-bot-dev:
@ravisiddhu

comment-bot-dev:
@ravisiddhu
Thanks for the PR! 🚀
- ⚠️check_terraform
 ```bash 
Running terraform fmt
modules/mysql/outputs.tf
--- old/modules/mysql/outputs.tf
+++ new/modules/mysql/outputs.tf
@@ -96,8 +96,8 @@
   description = "List of maps of additional users and passwords"
   value = [for r in google_sql_user.additional_users :
     {
-      name            = r.name
-      password        = r.password
+      name     = r.name
+      password = r.password
     }
   ]
   sensitive = true
modules/postgresql/outputs.tf
--- old/modules/postgresql/outputs.tf
+++ new/modules/postgresql/outputs.tf
@@ -101,8 +101,8 @@
   description = "List of maps of additional users and passwords"
   value = [for r in google_sql_user.additional_users :
     {
-      name            = r.name
-      password        = r.password
+      name     = r.name
+      password = r.password
     }
   ]
   sensitive = true
modules/mssql/outputs.tf
--- old/modules/mssql/outputs.tf
+++ new/modules/mssql/outputs.tf
@@ -65,8 +65,8 @@
   description = "List of maps of additional users and passwords"
   value = [for r in google_sql_user.additional_users :
     {
-      name            = r.name
-      password        = r.password
+      name     = r.name
+      password = r.password
     }
   ]
   sensitive = true
 ```

comment-bot-dev:
@ravisiddhu

comment-bot-dev:
@ravisiddhu

comment-bot-dev:
@ravisiddhu

comment-bot-dev:
@ravisiddhu

comment-bot-dev:
@ravisiddhu

comment-bot-dev:
@ravisiddhu
Thanks for the PR! 🚀
- ⚠️check_whitespace
 Failed whitespace check. More details below. 
 ```bash 
Checking for trailing whitespace
./docs/upgrading_to_sql_db_14.0.0.md:125:  
./docs/upgrading_to_sql_db_14.0.0.md:137:+     random_password = false 
./docs/upgrading_to_sql_db_14.0.0.md:139:  ] 
./docs/upgrading_to_sql_db_14.0.0.md:151:  
./docs/upgrading_to_sql_db_14.0.0.md:163:+     random_password = false 
./docs/upgrading_to_sql_db_14.0.0.md:165:  ] 
./docs/upgrading_to_sql_db_14.0.0.md:177:  
./docs/upgrading_to_sql_db_14.0.0.md:191:+     type            = null 
./docs/upgrading_to_sql_db_14.0.0.md:193:  ] 
./docs/upgrading_to_sql_db_14.0.0.md:205:  
./docs/upgrading_to_sql_db_14.0.0.md:219:+     random_password = false 
./docs/upgrading_to_sql_db_14.0.0.md:221:  ] 
Error: Trailing whitespace found in the lines above.
Checking for missing newline at end of file
Error: No newline at end of file ./docs/upgrading_to_sql_db_14.0.0.md
 ```

comment-bot-dev:
@ravisiddhu

bharathkkb (approved):

Issue #416, opened 2023-02-01
Error waiting for Create Instance: timeout while waiting for state to become 'DONE'


rohitjha941:
### TL;DR

Unable to create db instance, Timed out createation

### Expected behavior

DB should be created automatically

### Observed behavior

Error: Error waiting for Create Instance: timeout while waiting for state to become 'DONE' (last state: 'RUNNING', timeout: 15m0s)

### Terraform Configuration

```hcl
resource "random_password" "password" {
  length  = 16
  special = false
}

module "sq-private-service-access" {
  source      = "GoogleCloudPlatform/sql-db/google//modules/private_service_access"
  project_id  = var.project_id
  vpc_network = module.gcp-network.network_name
}

module "non-db" {
  source     = "GoogleCloudPlatform/sql-db/google//modules/safer_mysql"
  name       = var.name
  project_id = var.project_id

  deletion_protection  = false
  random_instance_name = true

  database_version  = "MYSQL_8_0"
  region            = var.region
  tier              = "db-f1-micro"
  zone              = "${var.region}-a"
  availability_type = "ZONAL"

  additional_users = [
    {
      name     = var.name
      password = random_password.password.result,
      type     = "BUILT_IN"
      host     = "0.0.0.0"
    }
  ]

  assign_public_ip   = "false"
  vpc_network        = module.gcp-network.network_self_link
  allocated_ip_range = module.sq-private-service-access.google_compute_global_address_name
  module_depends_on  = [module.sq-private-service-access.peering_completed]
}
```

### Terraform Version

```sh
Terraform v1.3.6
on linux_amd64
+ provider registry.terraform.io/gavinbunney/kubectl v1.14.0
+ provider registry.terraform.io/hashicorp/google v4.50.0
+ provider registry.terraform.io/hashicorp/helm v2.8.0
+ provider registry.terraform.io/hashicorp/kubernetes v2.17.0

Your version of Terraform is out of date! The latest version
is 1.3.7. You can update by downloading from https://www.terraform.io/downloads.html
```

philip-harvey:
Instances with <1 core take a LONG time to create. db-f1-micro will take ~25 minutes.  
The default in the module should be set to 30 minutes since that is what Terraform defaults to outside of using this module, but as a short term work around to this bug in the module try setting 
create_timeout = "30m"

PR #425, opened 2023-02-21
feat: increased timeout to 30m to support micro instances also.

ravisiddhu:
This PR is in response to https://github.com/terraform-google-modules/terraform-google-sql-db/issues/416
Buganizer issue : b/270095290

Instances with <1 core take a LONG time to create. db-f1-micro will take ~25 minutes. The default in the module should be set to 30 minutes since that is what Terraform defaults to outside of using this module. Hence setting the timeout as 30 minutes for CUD operations in all instance modules.


apeabody (approved):
Thanks for the contribution @ravisiddhu!

PR #427, opened 2023-02-22
feat: Set prevent_destroy = true for default database as a standard/default (MySQL)

zculek-fb:
Hello everyone,

I would like to add the option to enable or disable the _prevent_destroy_ lifecycle policy for the default database (MySQL).

At the moment, we are using the _deletion_protection_ option to prevent the deletion of the Cloud SQL MySQL instance, but we had a case when a series of unfortunate events caused Terraform to recreate our Cloud SQL instance and even though _deletion_protection_ kicked in and prevented the destruction of the instance, the instance database was still deleted/destroyed (just as the other instance resources), due to how destroy and deletion protection work (from what I gathered).

By being able to set _prevent_destroy_ lifecycle policy on the default database we would gain another layer of protection against loosing the data, which is more important to us than an instance being destroyed/deleted.

Of course, we had both internal and external backups and were able to recover, but this would prevent the need to go through the recovery process, which can be lengthy (depending on the DB size).

I've checked if this thing has already been proposed or discussed since I think a lot of people would benefit from this, but couldn't find anything related.

Thank you!


apeabody:
Thanks for the contribution @zculek-fb!  From the LINT:

```
Checking for documentation generation
diff -r '--exclude=.terraform' '--exclude=.kitchen' '--exclude=autogen' '--exclude=*.tfvars' '--exclude=*metadata.yaml' /workspace/modules/mssql/README.md /tmp/tmp.Mzkj41jv1b/generate_docs/workspace/modules/mssql/README.md
18c18
< | create\_timeout | The optional timeout that is applied to limit long database creates. | `string` | `"15m"` | no |
---
> | create\_timeout | The optional timeout that is applied to limit long database creates. | `string` | `"30m"` | no |
48c48
< | update\_timeout | The optional timeout that is applied to limit long database updates. | `string` | `"15m"` | no |
---
> | update\_timeout | The optional timeout that is applied to limit long database updates. | `string` | `"30m"` | no |
diff -r '--exclude=.terraform' '--exclude=.kitchen' '--exclude=autogen' '--exclude=*.tfvars' '--exclude=*metadata.yaml' /workspace/modules/mysql/README.md /tmp/tmp.Mzkj41jv1b/generate_docs/workspace/modules/mysql/README.md
15c15
< | create\_timeout | The optional timout that is applied to limit long database creates. | `string` | `"10m"` | no |
---
> | create\_timeout | The optional timout that is applied to limit long database creates. | `string` | `"30m"` | no |
21,22c21,22
< | db\_prevent\_destroy | Enable or disable prevent_destroy for default database | `bool` | `true` | no |
< | delete\_timeout | The optional timout that is applied to limit long database deletes. | `string` | `"10m"` | no |
---
> | db\_prevent\_destroy | Enable or disable prevent\_destroy on default database | `bool` | `true` | no |
> | delete\_timeout | The optional timout that is applied to limit long database deletes. | `string` | `"30m"` | no |
54c54
< | update\_timeout | The optional timout that is applied to limit long database updates. | `string` | `"10m"` | no |
---
> | update\_timeout | The optional timout that is applied to limit long database updates. | `string` | `"30m"` | no |
diff -r '--exclude=.terraform' '--exclude=.kitchen' '--exclude=autogen' '--exclude=*.tfvars' '--exclude=*metadata.yaml' /workspace/modules/postgresql/README.md /tmp/tmp.Mzkj41jv1b/generate_docs/workspace/modules/postgresql/README.md
15c15
< | create\_timeout | The optional timout that is applied to limit long database creates. | `string` | `"15m"` | no |
---
> | create\_timeout | The optional timout that is applied to limit long database creates. | `string` | `"30m"` | no |
22c22
< | delete\_timeout | The optional timout that is applied to limit long database deletes. | `string` | `"15m"` | no |
---
> | delete\_timeout | The optional timout that is applied to limit long database deletes. | `string` | `"30m"` | no |
53c53
< | update\_timeout | The optional timout that is applied to limit long database updates. | `string` | `"15m"` | no |
---
> | update\_timeout | The optional timout that is applied to limit long database updates. | `string` | `"30m"` | no |
diff -r '--exclude=.terraform' '--exclude=.kitchen' '--exclude=autogen' '--exclude=*.tfvars' '--exclude=*metadata.yaml' /workspace/modules/safer_mysql/README.md /tmp/tmp.Mzkj41jv1b/generate_docs/workspace/modules/safer_mysql/README.md
174c174
< | create\_timeout | The optional timout that is applied to limit long database creates. | `string` | `"15m"` | no |
---
> | create\_timeout | The optional timout that is applied to limit long database creates. | `string` | `"30m"` | no |
180c180
< | delete\_timeout | The optional timout that is applied to limit long database deletes. | `string` | `"15m"` | no |
---
> | delete\_timeout | The optional timout that is applied to limit long database deletes. | `string` | `"30m"` | no |
205c205
< | update\_timeout | The optional timout that is applied to limit long database updates. | `string` | `"15m"` | no |
---
> | update\_timeout | The optional timout that is applied to limit long database updates. | `string` | `"30m"` | no |
Error: Documentation generation has not been run, please run the
'make docker_generate_docs' command and commit the above changes.
Checking for trailing whitespace
Checking for missing newline at end of file
Running shellcheck
Checking file headers
Running flake8
/usr/local/bin/task_helper_functions.sh: line 30: DELETE_AT_EXIT: readonly variable
Checking for tflint
Working in ./examples/mssql-public ...
tflint passed ./examples/mssql-public 
Working in ./examples/mysql-backup-create-service-account ...
tflint passed ./examples/mysql-backup-create-service-account 
Working in ./examples/mysql-ha ...
tflint passed ./examples/mysql-ha 
Working in ./examples/mysql-private ...
tflint passed ./examples/mysql-private 
Working in ./examples/mysql-public ...
tflint passed ./examples/mysql-public 
Working in ./examples/postgresql-backup-provided-service-account ...
tflint passed ./examples/postgresql-backup-provided-service-account 
Working in ./examples/postgresql-ha ...
tflint passed ./examples/postgresql-ha 
Working in ./examples/postgresql-public ...
tflint passed ./examples/postgresql-public 
Working in ./examples/postgresql-public-iam ...
tflint passed ./examples/postgresql-public-iam 
Working in ./modules/backup ...
tflint passed ./modules/backup 
Working in ./modules/mssql ...
tflint passed ./modules/mssql 
Working in ./modules/mysql ...
tflint passed ./modules/mysql 
Working in ./modules/postgresql ...
tflint passed ./modules/postgresql 
Working in ./modules/private_service_access ...
tflint passed ./modules/private_service_access 
Working in ./modules/restore ...
tflint passed ./modules/restore 
Working in ./modules/safer_mysql ...
tflint passed ./modules/safer_mysql 
Skipping ./test/fixtures/mssql-ha
Skipping ./test/fixtures/mssql-public
Skipping ./test/fixtures/mysql-ha
Skipping ./test/fixtures/mysql-private
Skipping ./test/fixtures/mysql-public
Skipping ./test/fixtures/postgresql-ha
Skipping ./test/fixtures/postgresql-public
Skipping ./test/fixtures/postgresql-public-iam
Skipping ./test/setup
Running terraform fmt
modules/mysql/main.tf
--- old/modules/mysql/main.tf
+++ new/modules/mysql/main.tf
@@ -164,12 +164,12 @@
 }
 
 resource "google_sql_database" "default" {
-  count      = var.enable_default_db ? 1 : 0
-  name       = var.db_name
-  project    = var.project_id
-  instance   = google_sql_database_instance.default.name
-  charset    = var.db_charset
-  collation  = var.db_collation
+  count     = var.enable_default_db ? 1 : 0
+  name      = var.db_name
+  project   = var.project_id
+  instance  = google_sql_database_instance.default.name
+  charset   = var.db_charset
+  collation = var.db_collation
 
   lifecycle {
     prevent_destroy = var.db_prevent_destroy
```

zculek-fb:
@apeabody Thanks for the review and gcbrun. I've fixed some stuff, that should be it for now. Please let me know if you have any additional thoughts or comments on this.

apeabody:
Hi @zculek-fb! From the LINT:

```
There are some problems with the configuration, described below.

The Terraform configuration must be valid before initialization so that
Terraform can determine which modules and providers need to be installed.
╷
│ Error: Variables not allowed
│ 
│   on ../../modules/mysql/main.tf line 175, in resource "google_sql_database" "default":
│  175:     prevent_destroy = var.db_prevent_destroy
│ 
│ Variables may not be used here.
╵

╷
│ Error: Unsuitable value type
│ 
│   on ../../modules/mysql/main.tf line 175, in resource "google_sql_database" "default":
│  175:     prevent_destroy = var.db_prevent_destroy
│ 
│ Unsuitable value: value must be known
╵
```

zculek-fb:
So, I did some additional research on this and it seems that my initial approach won't work:

> Literal Values Only
> The lifecycle settings all affect how Terraform constructs and traverses the dependency graph.
> As a result, only literal values can be used because the processing happens too early for arbitrary expression evaluation.
(https://developer.hashicorp.com/terraform/language/meta-arguments/lifecycle#literal-values-only)

I also found out that there is quite an amount of people interested in changing this behavior for lifecycle policy, but it seems that the change can't be implemented easily and would require quite a rework on Terraform:

https://github.com/hashicorp/terraform/issues/4149
https://github.com/hashicorp/terraform/issues/30937
https://github.com/hashicorp/terraform/issues/22544

Doing some thinking and re-evaluation, I decided to go with a bit different approach - setting the prevent_destroy lifecycle option to true for default database as a new standard/default. Why would it make sense? Because:

1) You want to protect your data(base) as much as possible.
2) Default database is optional, not mandatory.
3) You can still delete the database via Console/SQL/CLI, this is only to prevent the database to be destroyed through Terraform.

So, it would still be possible to delete/remove the default database if you don't need it any more or want to create a new one, but you still have a bit more protection for the existing one.

What are your thoughts on this?

zculek-fb:
@apeabody Can you please run the checks again? Thanks!

@everyone: Any comments or suggestions on this "feature"?

g-awmalik:
@zculek-fb - thanks for the PR. It's fair to ask for  delete prevention on an instance but this should be configurable i.e. we should introduce a variable that can be passed into the module as opposed to hardcoding it to `true`.

My suggestion is:
- create a bool type variable to set prevention on/off
- create a simple example that verifies that behavior. Although for the test, you should skip the `teardown` stage since that won't be possible due to the deletion prevention.

g-awmalik (changes requested):
Please review my comment regarding making `prevent_destroy` configurable.

zculek-fb:
Hello @g-awmalik,

Thank you for checking into this and for your comments and suggestions. What you suggested was actually my initial intention, but unfortunately, it doesn't look to be possible - when I tried to set the variable instead of hardcoding it to true, I got this error:

│ Error: Variables not allowed
│ 
│   on ../../modules/mysql/main.tf line 175, in resource "google_sql_database" "default":
│  175:     prevent_destroy = var.db_prevent_destroy
│ 
│ Variables may not be used here.

So I tried to think of something else. And as I wrote - this is not the instance deletion prevention, this is just for the default database of the primary instance, and it will only work in the cases when something will happen that will cause Terraform to try to delete the database, which is actually/exactly what I want to prevent, because even if you use instance deletion protection (deletion_protection and/or deletion_protection.enabled), the instance itself will be protected from deletion, but in the eventual process, the database will be deleted as a resource before the deletion_protection of the instances kicks in (which happened to us), and so the empty instance will be saved from the deletion, but the data lost. Adding prevent_destroy to default database should save people from database being deleted by Terraform unintentionally and from having to restore the DB from backup.

It would still be possible to delete the default database manually (e.g. trough UI or mysql cli), this won't interfere.

But, if there is a way to set this as an option/variable instead of hardcoding it, I am all for it, as that would be much better solution. Maybe you or someone know of one?

Thanks!

zculek-fb:
Let me add an example of what I'm talking about. I created a TF plan which involves a destroy/destructive operation on a Cloud SQL instance with both `deletion_protection` and `deletion_protection.enabled` set to `true`. When the plan is applied, instance destruction is prevented, but due to the how the process goes, database(s) and user(s) are destroyed prior to deletion protection kicks in:

`Do you want to perform these actions?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: yes
module.mysql_db.google_sql_user.default[0]: Destroying... [id=sascsql//sascsql]
module.mysql_db.google_sql_database.default[0]: Destroying... [id=projects/freshbooks-staging/instances/sascsql/databases/sascsql]
module.mysql_db.google_sql_user.default[0]: Destruction complete after 1s
module.mysql_db.random_password.user-password: Destroying... [id=none]
module.mysql_db.google_sql_database_instance.replicas["sascsql-replica"]: Destroying... [id=sascsql-read-replica]
module.mysql_db.random_password.user-password: Destruction complete after 0s
module.mysql_db.google_sql_database.default[0]: Destruction complete after 2s
╷
│ Error: Error, failed to delete instance because deletion_protection is set to true. Set it to false to proceed with instance deletion
│
│`

I guess I understand the logic here - you must first destroy instance resources (parts) and then destroy the instance, but I wonder why some kind of IF logic wasn't implemented to first check if deletion_protection is enabled, and if it is, skip the destruction process as a whole, instead of destroying resources which are part of the instance and then leaving the instance alive in the end.

From my point of view (and experience), it's much easier and faster to re-create a Cloud SQL instance (few mins) than to have to restore a database from a backup which was deleted this way, since databases could be hundreds of GBs in size, and the restore, no matter on the type, takes a quite amount of time.

g-awmalik:
@zculek-fb - we have a variable called [`database_deletion_policy`](https://github.com/terraform-google-modules/terraform-google-sql-db/blob/master/modules/postgresql/variables.tf#L377) that can be set on all database modules. Setting that to "ABANDON" should do that trick. Have you looked at that?

rileykarson:
@zculek-fb: Terraform Core calls the provider's delete method on the resource but the implementation of that delete method performs a no-op: https://github.com/hashicorp/terraform-provider-google-beta/blob/07415a4f8718a760f17680b758aa526fbd433288/google-beta/resource_sql_database.go#L356-L360

zculek-fb:
Thank you for the info @rileykarson!

OK, I guess this way of getting the database protected from deletion/destruction inside a module or template isn't the best solution and using database_deletion_policy = ABANDON would actually be a pretty dirty workaround rather than the real solution, so I'm going to close this one out and try a different approach on this matter.

Thanks everyone for your comments, checks and suggestions, they have been very helpful!

Issue #438, opened 2023-03-10
Allow users to configure connector enforcement for Mysql Databases


maguec:
### TL;DR

Allow customers to set a variable to modify the value for connector_enforcement

### Terraform Resources

```markdown
https://cloud.google.com/sql/docs/mysql/admin-api/rest/v1/instances#ConnectorEnforcement
```

### Detailed design

```markdown
If the variable is set to true then set  "REQUIRED" else set as the default setting of "NOT_REQUIRED"
```

PR #439, opened 2023-03-10
feat: make connector enforcement configurable

maguec:
Fixes https://github.com/terraform-google-modules/terraform-google-sql-db/issues/438

Allows Mysql module users the ability to modify the connector_enforcement setting


modules/mysql/metadata.yaml
```diff
@@ -100,6 +100,11 @@ spec:
       start_time: null
       transaction_log_retention_days: null
     required: false
+  - name: connector_enforcement
+    description: Force the client to use the connector. See [more details](https://cloud.google.com/sql/docs/mysql/connect-connectors)
+    type: bool
+    default: true
```
isaurabhuttam, line 106: Let's change the default value as false, this is what we want as per documentation and variable file.

modules/mysql/README.md
isaurabhuttam, line 1: We need to change our commit message to follow format mentioned here: https://www.conventionalcommits.org/en/v1.0.0/

isaurabhuttam:
We need to change our commit message to follow format mentioned here: https://www.conventionalcommits.org/en/v1.0.0/

modules/mysql/metadata.yaml
maguec: nice catch.  Updated

isaurabhuttam:
Can you please pull latest main in your local and merge this branch with latest local main? this should remove `This branch is out-of-date with the base branch` message. Before pulling latest master, make sure you sync your forked repository.

g-awmalik (approved):
LGTM. Please take a look at the var suggestion.

modules/mysql/main.tf
```diff
@@ -55,6 +58,7 @@ resource "google_sql_database_instance" "default" {
     activation_policy           = var.activation_policy
     availability_type           = var.availability_type
     deletion_protection_enabled = var.deletion_protection_enabled
+    connector_enforcement       = local.connector_enforcement
```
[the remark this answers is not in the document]
maguec: it is not True or False in the API, so using a local will allow us to convert the boolean to the necessary values hiding the complexity from the customer.

```Option values include the following: NOT_REQUIRED (Cloud SQL instances can be connected without Cloud SQL Connectors) and REQUIRED (Only allow connections that use Cloud SQL Connectors)```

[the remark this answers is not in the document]
g-awmalik: The usage of locals is only useful when 
- the local is being used in multiple places
- the logic is fairly complex for the attribute being set

None of these apply here and IMO it will actually increase readability for the consumer since they'll be able understand how this attribute is set without having to scroll back and forth.

Either way its not a huge deal; we try to make things as readable as possible with all modules, hence the comment. Feel free to resolve this and merge if you don't want to do this right now.

PR #440, opened 2023-03-12
docs: `make generate_docs`

mnahkies:
run doc generation to pick-up default value change from #425

I think that ideally this would be checked for by CI


apeabody:
> I think that ideally this would be checked for by CI

Hi @mnahkies, the LINT does check for documentation generateion.  Looking at the example the `check_documentation` step did fail, however for some reason the overall LINT status was returned to GitHub as pass?

@bharathkkb example: https://github.com/terraform-google-modules/terraform-google-sql-db/pull/425/checks?check_run_id=11495645443

apeabody (approved):
Thanks for the contribution @mnahkies!

I confirmed in the LINT details everything passed.

Issue #445, opened 2023-03-20
Hardcoded values forcing random_password recreation


kpocius:
### TL;DR

Much like https://github.com/terraform-google-modules/terraform-google-sql-db/issues/418 there's a new issue which forces recreating all random passwords. This time it's cause by `min_lower`, `min_upper`, and `min_numeric` being hardcoded. The change was introduced in https://github.com/terraform-google-modules/terraform-google-sql-db/pull/443

### Expected behavior

Updating the module without any other changes should not cause passwords to be recreated.

### Observed behavior

All random passwords are being recreated due to changes in requirements:

```

  # module.postgresql.random_password.user-password must be replaced
-/+ resource "random_password" "user-password" {
!       bcrypt_hash = (sensitive value)
!       id          = "none" -> (known after apply)
!       min_lower   = 0 -> 1 # forces replacement
!       min_numeric = 0 -> 1 # forces replacement
!       min_upper   = 0 -> 1 # forces replacement
!       result      = (sensitive value)
        # (8 unchanged attributes hidden)
    }
```

### Terraform Configuration

```hcl
N/A
```

### Terraform Version

```sh
N/A
```

kpocius:
@ravisiddhu, any chance you could handle this?

ravisiddhu:
Yes, I will work on it.

PR #446, opened 2023-03-21
fix: prevent auto-update of random_password resource

ravisiddhu:
This is to fix https://github.com/terraform-google-modules/terraform-google-sql-db/issues/445

```
fix: prevent update of random_password resource due to the changes in the hardcoded values with the help of lifecycles.
```


kpocius:
I can confirm this indeed resolves #445

g-awmalik (approved):

Issue #455, opened 2023-04-22
Add map of IAM users to module output


kriswuollett:
### TL;DR

A map of IAM database users should be provided in a database cluster module output so that the actual username is available downstream which is different than the email when it is a service account because its `.gserviceaccount.com` suffix gets trimmed. This output would be useful for downstream Terraform modules, e.g., passing the username to a `google_cloud_run_v2_job` that uses Cloud SQL Proxy via `cloud_sql_instance` volume.

### Terraform Resources

- https://registry.terraform.io/providers/hashicorp/google/latest/docs

### Detailed design

In the `postgres` module for example, the `iam_users` is defined as the following which is also directly used as an output:

https://github.com/terraform-google-modules/terraform-google-sql-db/blob/15298c275e2d531e3379d1c445b2317c99e6363b/modules/postgresql/main.tf#L29

Following that something like the following should be added:

```terraform
  iam_user_accounts = {
    for id, user in local.iam_users id => {
      email         = user.email,
      is_account_sa = user.is_account_sa
      username = user.is_account_sa ? trimsuffix(user.email, ".gserviceaccount.com") : user.email
    }
  }
```

Then to `outputs.tf`:

```terraform
output "iam_users_map" {
  description = "The map of the IAM users with access to the CloudSQL instance"
  value       = local.iam_user_accounts
}
```

### Additional information

This feature would be similar to the map provided by the `terraform-google-modules/service-accounts/google`: 

https://github.com/terraform-google-modules/terraform-google-service-accounts/blob/7c9e85d6d7b16a30da1451c0ce65d0e64016cacb/outputs.tf#L43-L46

This module already asks for an `id` that doesn't appear to be used for anything other than a key for a for expression, so it seemed like the intent would be to support such as the map described, otherwise the database username itself would have been a sufficient key for the `google_sql_user` for expression.

PR #458, opened 2023-04-23
feat: add `google_sql_user` `name` to `iam_users` output

### LINKED ISSUE #456: potential typo in `postgres` user service account email detection
### TL;DR

The is service account check in the `postgres` module doesn't appear to be technically correct because something like `oscar@notgserviceaccount.com` would be detected as a service account, and which email is possible to be created because it is not the `gserviceaccount.com` domain:

https://github.com/terraform-google-modules/terraform-google-sql-db/blob/15298c275e2d531e3379d1c445b2317c99e6363b/modules/postgresql/main.tf#L32

Perhaps it was a typo because further down in the code a leading `.` is used:

https://github.com/terraform-google-modules/terraform-google-sql-db/blob/15298c275e2d531e3379d1c445b2317c99e6363b/modules/postgresql/main.tf#L270-L274

Not the place where it may be necessary, but technically also checking for proper [Project ID](https://cloud.google.com/resource-manager/docs/creating-managing-projects) could be done with a regex which I think the following would be sufficient:

`@[a-z][a-z0-9-]{4,28}[a-z0-9]\.iam\.gserviceaccount\.com$`.

I assume applying the terraform with an email account like `oscar@notgserviceaccount.com` would fail anyways because I assume GCP wouldn't find that service account.

### Expected behavior

_No response_

### Observed behavior

_No response_

### Terraform Configuration

```hcl
N/A
```

### Terraform Version

```sh
N/A
```

### Additional information

_No response_

### LINKED ISSUE #457: `iam_users` output not part of dependency graph
### TL;DR

The `iam_users` output does not follow the [best practice of it being able to be part of the dependency graph](https://cloud.google.com/docs/terraform/best-practices-for-terraform#outputs).

It should include something calculated from a resource.

### Expected behavior

_No response_

### Observed behavior

_No response_

### Terraform Configuration

```hcl
N/A
```

### Terraform Version

```sh
N/A
```

### Additional information

_No response_

kriswuollett:
Added `name` to `iam_users` output so that downstream logic does not need to repeat the IAM service account SQL username calculation (partially addresses #455, fixes #456). 

Also ensured that `iam_users` output creates an implicit dependency on the relevant `google_sql_user` resource (fixes #457). And also updated `iam_users` output docs to not imply that `cloudsql.instances.login` IAM role is also granted.

Replaced internal usage of `is_account_sa` boolean with `type` from `google_sql_user` enum string as it is theoretically possible that another type could be added like `CLOUD_IAM_GROUP`.

### DIFF
--- a/modules/postgresql/README.md
+++ b/modules/postgresql/README.md
@@ -32,7 +32,7 @@ Note: CloudSQL provides [disk autoresize](https://cloud.google.com/sql/docs/mysq
 | enable\_random\_password\_special | Enable special characters in generated random passwords. | `bool` | `false` | no |
 | encryption\_key\_name | The full path to the encryption key used for the CMEK disk encryption | `string` | `null` | no |
 | follow\_gae\_application | A Google App Engine application whose zone to remain in. Must be in the same region as this instance. | `string` | `null` | no |
-| iam\_users | A list of IAM users to be created in your CloudSQL instance | <pre>list(object({<br>    id    = string,<br>    email = string<br>  }))</pre> | `[]` | no |
+| iam\_users | A list of IAM account emails to create corresponding CloudSQL instance users | <pre>list(object({<br>    id    = string,<br>    email = string<br>  }))</pre> | `[]` | no |
 | insights\_config | The insights\_config settings for the database. | <pre>object({<br>    query_string_length     = number<br>    record_application_tags = bool<br>    record_client_address   = bool<br>  })</pre> | `null` | no |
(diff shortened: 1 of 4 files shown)


g-awmalik:
Can you rebase this with master please?

kriswuollett:
@g-awmalik done

g-awmalik:
@kriswuollett - can you please rebase (I can't since this seems to have been forked in an org) and run `make generate_docs` as well? Thanks.

PR #461, opened 2023-04-27
Fix to avoid password reset for existing additional users

Anand-chenepalli:
- Fix to avoid password reset for existing additional users


Anand-chenepalli:
@g-awmalik Accepted CLA  and please let me know if any further steps need to be done from my end

Issue #482, opened 2023-06-29
Allow configuration of query insights sampling rate


jawnsy:
### TL;DR

The provider supports configuring the [sample rate for Query Insights](https://cloud.google.com/sql/docs/postgres/using-query-insights#enable-insights), and the module should allow it, too.

### Terraform Resources

[query_plans_per_minute](https://registry.terraform.io/providers/hashicorp/google/4.71.0/docs/resources/sql_database_instance#query_plans_per_minute) - Number of query execution plans captured by Insights per minute for all queries combined. Between 0 and 20. Default to 5.

### Detailed design

Looks like a straightforward addition here: https://github.com/terraform-google-modules/terraform-google-sql-db/blob/2fcad8c54193a4c564b26d4bd8fba2b7ec788b5c/modules/postgresql/main.tf#L110-L119

It may be a backward incompatible change because it requires changing the variable type definition: https://github.com/terraform-google-modules/terraform-google-sql-db/blob/2fcad8c54193a4c564b26d4bd8fba2b7ec788b5c/modules/postgresql/variables.tf#L193-L201

It looks like this field was introduced in v4.44.0 of the provider via https://github.com/hashicorp/terraform-provider-google/pull/12951, see notes here https://github.com/hashicorp/terraform-provider-google/releases/tag/v4.44.0

PR #484, opened 2023-06-29
feat: add support for query_plans_per_minute

jawnsy:
This adds support for configuring the Query Insights sampling rate via the query_plans_per-minute setting.

Closes: #482


kam1kaze:
What do we need to merge such feature?

g-awmalik (approved):

PR #479, opened 2023-06-23
feat: Added new time_zone variable for MSSQL

Anand-chenepalli:
Added new time_zone variable for MSSQL

PR #488, opened 2023-07-17
feat: add time_zone for mssql

vponnam:
This is the same change being submitted through https://github.com/terraform-google-modules/terraform-google-sql-db/pull/479, submitting a new PR to address the linter check failures on the other PR.

cc @Anand-chenepalli

### DIFF
--- a/modules/mssql/variables.tf
+++ b/modules/mssql/variables.tf
@@ -306,3 +306,9 @@ variable "deletion_protection" {
   type        = bool
   default     = true
 }
+
+variable "time_zone" {
+  description = "The time zone for SQL instance."
+  type        = string
+  default     = "UTC"
+}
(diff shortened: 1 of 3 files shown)


modules/mssql/variables.tf
```diff
@@ -306,3 +306,9 @@ variable "deletion_protection" {
   type        = bool
   default     = true
 }
+
+variable "time_zone" {
+  description = "The time zone for SQL instance."
+  type        = string
+  default     = "UTC"
```
bharathkkb, line 313: nit: can we set this to null so we use whatever is provider default?

bharathkkb (approved):

modules/mssql/variables.tf
vponnam: > nit: can we set this to null so we use whatever is provider default?

Nice catch, it was a bit of an overlook on my end, but fixed now. Thanks @bharathkkb

PR #491, opened 2023-07-19
feat!: added `edition` in mssql, mysql, postgresql, safer_sql and `data_cache_config` to mysql and safer_sql modules

ibhaskar2:
b/291924307


modules/mysql/variables.tf
```diff
@@ -161,6 +167,14 @@ variable "user_labels" {
   description = "The key/value labels for the master instances."
 }
 
+variable "data_cache_config" {
```
imrannayer, line 170: Since there is only one value can you just change it to string variable:

```
variable "data_cache_enabled" {
  description = "Whether data cache is enabled for the instance. Defaults to false. Feature is only available for ENTERPRISE_PLUS tier and supported database_versions"
  type        = bool
  default     = false
}
```

modules/mysql/main.tf
```diff
@@ -94,6 +95,12 @@ resource "google_sql_database_instance" "default" {
         record_client_address   = lookup(insights_config.value, "record_client_address", false)
       }
     }
+    dynamic "data_cache_config" {
```
imrannayer, line 98: @ibhaskar2 plz make following change otherwise tests will fail. This value is immutable only for ENTERPRISE_PLUS tier and when setting value to true. Otherwise it will keep updating the value. 

```
    dynamic "data_cache_config" {
      for_each = var.edition == "ENTERPRISE_PLUS" && var.data_cache_enabled ? ["cache_enabled"] : []
      content {
        data_cache_enabled = var.data_cache_enabled
      }
    }
```

imrannayer:
@ibhaskar2 you need to change min provider version to [4.74](https://github.com/hashicorp/terraform-provider-google/blob/main/CHANGELOG.md#4740-july-18-2023) as these features were part of the release.

modules/mysql/variables.tf
ibhaskar2: Done

modules/mysql/read_replica.tf
```diff
@@ -37,6 +37,7 @@ resource "google_sql_database_instance" "replicas" {
 
   settings {
     tier                        = lookup(each.value, "tier", var.tier)
+    edition                     = lookup(each.value, "edition", var.edition)
```
isaurabhuttam, line 40: is `data_cache_config` applicable to mysql read replica also?

ibhaskar2: Yes, verified it with mysql team and manually via creating replica with data cache on.

isaurabhuttam: so, we are adding the data_cache_config here also right?

ibhaskar2: yes

ibhaskar2:
@imrannayer can you review it and merge it?

imrannayer:
@bharathkkb can you plz review?

g-awmalik (approved):
LGTM.

g-awmalik:
@isaurabhuttam - since this is a breaking change can you please add an upgrade doc like [this](https://github.com/terraform-google-modules/terraform-google-sql-db/blob/master/docs/upgrading_to_sql_db_11.0.0.md). This should be pretty simple i.e. explaining customers to update version constraints, if they have any for TPG.

PR #497, opened 2023-08-08
DB must not exist before import for SQL Server

zeraholladay:
Database should not exist before SQL Server import runs.


g-awmalik:
@zeraholladay - do you mind explaining the issue in a bit more detail?

@ravi-dalal - FYI since you wrote this module.

PR #500, opened 2023-08-15
feat: config connector_enforcement in postgres

NissesSenap:


NissesSenap:
Just as mysql, us postgres users also want to be able to use connector enforcement, see https://github.com/terraform-google-modules/terraform-google-sql-db/pull/439/ for the original PR.

g-awmalik (approved):

PR #513, opened 2023-09-14
fix: Don't require `var.zone` for `read_replica` instances

fatmcgav:
Currently, a value is required for either `var.read_replicas[replica]["zone"]` or `var.zone` as a fallback.

This commit tweaks that behaviour to use _either_ `var.zone` or a random
available zone as the fallback.

Backwards compatibility should be retained, as `local.zone` already
prefers the value of `var.zone` over a random available zone.

Also worth noting that `local.zone` is already used in the
`location_preference {}` block.


g-awmalik:
@fatmcgav - sorry for the delay on this. Can you please sign the CLA?

fatmcgav:
@g-awmalik no worries... So I've changed jobs since opening this PR, but will see if can sort the cla out tomorrow 👍

Issue #214, opened 2021-04-30
Enable root_password for postgres


BusiPlay:
Per https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/sql_database_instance#root_password:

(Optional) Initial root password. Required for MS SQL Server, ignored by MySQL and PostgreSQL.

The creation of a CloudSQL Postgres instance automatically creates a postgres user, whose password can be set through API per https://cloud.google.com/sql/docs/postgres/create-manage-users#rest-v1beta4

Right now in order to create a root user via Terraform, a separate cloudsql_database_user resource must be created with root privileges.

Being able to set the postgres user password through the existing root_password attribute seems like a better approach, as it reduces the number of users being given administrative privileges to the database instance.

BusiPlay:
This statement applies only to MySQL 2nd Generation instances. Postgres instances do create a postgres user, this is not deleted by Terraform currently. This request was entered to allow the root_password attribute to update this default postgres user as it does for SQL Server.

imrannayer:
It still applies to provider not this module. Right now provider uses `root_password` paraeter for MsSQL instance only and ignores for MySQL and Postgres. You may wana create issue for on [provider's site](https://github.com/hashicorp/terraform-provider-google).
Once provider has the capability it can be added to module.

BusiPlay:
right, my error - I will direct this to the provider site. Thanks!

the-veloper:
I think the `google_sql_database_instance` resource from the provider can now accept `root_password` according to the official google docs:
https://cloud.google.com/sql/docs/postgres/samples/cloud-sql-postgres-instance-pvp

PR #521, opened 2023-10-05
feat(postgres): add root password parameter

the-veloper:
Apparently a compliant `root_password` value is required, if there's a password policy set, due to the fact that:
```
NOTE on google_sql_database_instance:
- Second-generation instances include a default 'root'@'%' user with no password. This user will be deleted by Terraform on instance creation. You should use google_sql_user to define a custom user with a restricted host and strong password.
```

Otherwise there's the following error:
```
 Error: Error, failed to create instance postgre-dev-01: googleapi: Error 400: Invalid request: Root user's password is not policy compliant: Length of password cannot be less than 8. ., invalid
│
│   with module.postgre.google_sql_database_instance.default,
│   on .terraform/modules/postgre/modules/postgresql/main.tf line 50, in resource "google_sql_database_instance" "default":
│   50: resource "google_sql_database_instance" "default" {
```

Note source: https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/sql_database_instance
Example from google: https://cloud.google.com/sql/docs/postgres/samples/cloud-sql-postgres-instance-pvp


imrannayer:
@the-veloper thanks for the PR. Lint test failed.
Thanks

g-awmalik:
This will addres #214

g-awmalik (approved):

PR #541, opened 2023-11-16
fix!: safer_mysql module's `assign_public_ip` input should be bool type

leavesster:
## Fix

the `safer_mysql` submodule's `assign_public_ip`'s type is `string`, but default value is `false` which is bool type

```
variable "assign_public_ip" {
  description = "Set to true if the master instance should also have a public IP (less secure)."
  type        = string
  default     = false
}
```

and this variable is used as bool parameter to `ip_configuration.ipv4_enabled`

https://github.com/terraform-google-modules/terraform-google-sql-db/blob/5bcd8a4824f3c021d81842e9c7045af517a15827/modules/safer_mysql/main.tf#L54

but the ipv4_enabled is require bool type not string.
It also effects example [mysql-private](https://github.com/terraform-google-modules/terraform-google-sql-db/tree/master/examples/mysql-private) at https://github.com/terraform-google-modules/terraform-google-sql-db/blob/5bcd8a4824f3c021d81842e9c7045af517a15827/examples/mysql-private/main.tf#L97

## current action

if I run `terraform apply` in [example/mysql-private](https://github.com/terraform-google-modules/terraform-google-sql-db/tree/master/examples/mysql-private) , it will reject to run because of this wrong type.


imrannayer:
@leavesster thanks for the PR. lint test is failing. Plz generate docs. You can follow [this doc](https://github.com/terraform-google-modules/terraform-google-sql-db/blob/master/CONTRIBUTING.md)

imrannayer (approved):

Issue #524, opened 2023-10-09
dns_name missing from postgresql module outputs


steven-edgar:
### TL;DR

Add dns_name to outputs of postgresql module

### Expected behavior

dns_name value outputted from postgresql module.
Note: dns_name is not in the documentation for the google_sql_database_instance terraform resource, this has been reported separately via issue (https://github.com/hashicorp/terraform-provider-google/issues/16176)

### Observed behavior

No dns_name output

### Terraform Configuration

```hcl
n/a
```

### Terraform Version

```sh
n/a
```

### Additional information

Append the following to modules/postgresql/outputs.tf :

`
output "dns_name" {
  value = module.postgresql-db.default.dns_name
  description = "DNS name of the instance endpoint"
}
`

PR #544, opened 2023-11-24
fix: add dns_name to output in postgresql module

dkuji:
resolve: #524


dkuji:
@isaurabhuttam @imrannayer Good Day! can I get a review?

imrannayer:
@dkuji thanks for the PR. Do we need this? We already have an output `primary` which has all the output of primary instance. You can grab DNS value from it.

g-awmalik (approved):

PR #575, opened 2024-02-06
feat!: Allow passing ssl_mode for MySQL Module

davidcollom:
There is currently no way of passing the `ssl_mode` value from the module to the `cloudsql_instance` resource.

This PR aims to address this issue.


davidcollom:
Hey @imrannayer thanks for your quick response.

As for postgresql I'm not sure, as I don't use this.

As for the `safer_mysql` module... Would you expect `TRUSTED_CLIENT_CERTIFICATE_REQUIRED` to be enforced? Or default? As this imo would be the "safest default value"..or would/should we fall back to the providers default?

davidcollom:
@imrannayer I've updated the `safer_mysql` module, and checked the postgresql module, which already includes this feature, which was added here: https://github.com/terraform-google-modules/terraform-google-sql-db/commit/9c59232a37725aef57162e836090c64495e6acd9

g-awmalik:
@imrannayer - are you good with the changes?

imrannayer:
LGTM

davidcollom:
Hey @imrannayer @g-awmalik  what's needed to get this merged in? All the CI has passed but pending code reviews from yourselves.

g-awmalik (approved):

Issue #552, opened 2023-12-12
Issue With SSL Mode on v18.0.0. change provider min version to 5.6 for postgres module.


GusPrice:
### TL;DR

Get the following error from Terraform when using the postgres module

```
Error: Unsupported argument
on .terraform/modules/postgres_db/modules/postgresql/main.tf line 108, in resource "google_sql_database_instance" "default":
        ssl_mode                                      = lookup(ip_configuration.value, "ssl_mode", null)
An argument named "ssl_mode" is not expected here.
```

### Expected behavior

Expect ssl_mode to be applied appropriately.

### Observed behavior

See error above.

### Terraform Configuration

```hcl
On module v18.0.0


  ip_configuration = {
    ipv4_enabled    = false
    private_network = data.google_compute_network.primary_network.id
    require_ssl     = true
    ssl_mode        = "TRUSTED_CLIENT_CERTIFICATE_REQUIRED"
  }
```
I also tried
```
  ip_configuration = {
    ipv4_enabled    = false
    private_network = data.google_compute_network.primary_network.id
    require_ssl     = false
    ssl_mode        = "TRUSTED_CLIENT_CERTIFICATE_REQUIRED"
  }
```

The rest of my configuration works without ssl_mode on v17.1.0
```


### Terraform Version

```sh
Terraform Version 1.6.5

    google = {
      version = "~> 4.84.0"
    }

    google-beta = {
      version = "~> 4.84.0"
    }
```

volver-13:
Seems like the module should require a minimum of v5.6.0 provider version

https://github.com/hashicorp/terraform-provider-google/releases/tag/v5.6.0

_sql: added ssl_mode field to google_sql_database_instance resource (https://github.com/hashicorp/terraform-provider-google/pull/16394)_

ykyr:
There was just [v19.0 released](https://github.com/terraform-google-modules/terraform-google-sql-db/releases), without change for the minimum google provider version :(

PR #577, opened 2024-02-09
fix: bump required google provider to 5.6 for postgres

ykyr:
Fixes https://github.com/terraform-google-modules/terraform-google-sql-db/issues/552

### DIFF
--- a/modules/postgresql/versions.tf
+++ b/modules/postgresql/versions.tf
@@ -27,11 +27,11 @@ terraform {
     }
     google = {
       source  = "hashicorp/google"
-      version = ">= 4.80.0, < 6"
+      version = ">= 5.6, < 6"
     }
     google-beta = {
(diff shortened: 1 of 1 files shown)


modules/postgresql/versions.tf
```diff
@@ -27,11 +27,11 @@ terraform {
     }
     google = {
       source  = "hashicorp/google"
-      version = ">= 4.80.0, < 6"
+      version = ">= 5.6.0, < 6"
```
imrannayer, line 30: can u pl change it to

```
version = ">= 5.6, < 6"
```

ykyr: ah, right. Fixed. 
Thanks!

imrannayer (approved):

apeabody:
Looks like this should be marked as a breaking change?

PR #591, opened 2024-04-11
feat: Add support for email based alerting

ps-occrp:
Add support for email based alerting in case on successful backups or exports within specified frequency.


ps-occrp:
@imrannayer can you please review this PR?

modules/backup/main.tf
```diff
@@ -52,14 +54,22 @@ data "google_sql_database_instance" "backup_instance" {
   project = var.project_id
 }
 
+resource "google_monitoring_notification_channel" "email" {
```
imrannayer, line 57: @ps-occrp is it possible to make channel name a variable?
Also can you add another bool variable create_email_notification_channel. If true then create channel otherwise expect channel is already created outside of this module.

ps-occrp: yes, good idea :)

imrannayer (approved):

Issue #599, opened 2024-04-24
backup stopped working because missing permission storage.objects.delete


rverenich:
https://github.com/terraform-google-modules/terraform-google-sql-db/blob/fc37d6e6a7c37625ea95770d386e4b3033926926/modules/backup/main.tf#L140-L145

received today
stopped working because this permission appeared to be needed

(previously `role   = "roles/storage.objectCreator"` was enough)

```{
  "protoPayload": {
    "@type": "type.googleapis.com/google.cloud.audit.AuditLog",
    "status": {
      "code": 7,
      "message": "gs://.../: Access denied for service account ...@gcp-sa-cloud-sql.iam.gserviceaccount.com, permissions missing are: [storage.objects.delete]"
    },
    "authenticationInfo": {
      "principalEmail": "...@...",
      "serviceAccountDelegationInfo": [
        {
          "firstPartyPrincipal": {
            "principalEmail": "...@..."
          }
        }
      ],
      "principalSubject": "serviceAccount:...@..."
    },
    "requestMetadata": {
      "callerIp": "...",
      "requestAttributes": {
        "time": "2024-04-24T04:30:00.879795Z",
        "auth": {}
      },
      "destinationAttributes": {}
    },
    "serviceName": "cloudsql.googleapis.com",
    "methodName": "cloudsql.instances.export",
    "authorizationInfo": [
      {
        "resource": "projects/...",
        "permission": "cloudsql.instances.export",
        "granted": true,
        "resourceAttributes": {
          "service": "sqladmin.googleapis.com",
          "name": "...",
          "type": "sqladmin.googleapis.com/Instance"
        },
        "permissionType": "DATA_READ"
      }
    ],
    "resourceName": "...",
    "request": {
      "instance": "...",
      "project": "...",
      "body": {
        "exportContext": {
          "uri": "gs://....sql.gz",
          "offload": false
        }
      },
      "@type": "type.googleapis.com/google.cloud.sql.v1.SqlInstancesExportRequest"
    }
  },
  "insertId": "...",
  "resource": {
    "type": "cloudsql_database",
    "labels": {
      "database_id": "...",
      "region": "europe-west1",
      "project_id": "..."
    }
  },
  "timestamp": "2024-04-24T04:30:00.705844Z",
  "severity": "ERROR",
  "logName": "projects/.../logs/cloudaudit.googleapis.com%2Fdata_access",
[...]

PR #600, opened 2024-04-24
fix: update bucket permissions

rverenich:
closest existing role to fix https://github.com/terraform-google-modules/terraform-google-sql-db/issues/599 and add [storage.objects.delete] permission is roles/storage.objectUser. 

https://www.googlecloudcommunity.com/gc/Databases/Export-with-gcloud-sql-export-but-service-role-with-storage/m-p/666282 quotes from answer from google staff:
To export data from Cloud SQL to Cloud Storage, the service account should have the roles/storage.legacyBucketWriter role on the bucket. Note: The roles/storage.legacyBucketWriter role is a legacy role and might not be recommended for all use cases. For more granular control, consider using roles like roles/storage.objectCreator or roles/storage.objectAdmin.

### DIFF
--- a/modules/backup/main.tf
+++ b/modules/backup/main.tf
@@ -178,6 +178,13 @@ resource "google_storage_bucket_iam_member" "sql_instance_account" {
   count  = var.enable_export_backup ? 1 : 0
   bucket = split("/", var.export_uri)[2] #Get the name of the bucket out of the URI
   member = "serviceAccount:${data.google_sql_database_instance.backup_instance.service_account_email_address}"
+  role   = "roles/storage.objectCreator"
+}
+
+resource "google_storage_bucket_iam_member" "sql_instance_account_objectuser" {
+  count  = var.enable_export_backup ? 1 : 0
+  bucket = split("/", var.export_uri)[2] #Get the name of the bucket out of the URI
+  member = "serviceAccount:${data.google_sql_database_instance.backup_instance.service_account_email_address}"
   role   = "roles/storage.objectUser"
 }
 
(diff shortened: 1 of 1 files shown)


g-awmalik:
@rverenich - thanks for the PR. Can you please sign the CLA?

rverenich:
signed

ps-occrp:
In current form this service account does not need to delete any objects from a bucket, why should it get `roles/storage.objectUser` role?

modules/backup/main.tf
```diff
@@ -178,7 +178,7 @@ resource "google_storage_bucket_iam_member" "sql_instance_account" {
   count  = var.enable_export_backup ? 1 : 0
   bucket = split("/", var.export_uri)[2] #Get the name of the bucket out of the URI
   member = "serviceAccount:${data.google_sql_database_instance.backup_instance.service_account_email_address}"
-  role   = "roles/storage.objectCreator"
+  role   = "roles/storage.objectUser"
```
g-awmalik, line 181: Instead of replacing `roles/storage.objectCreator` with `roles/storage.objectUser`, I would suggest adding another IAM binding for `roles/storage.objectViewer` to give the SA read permissions.

g-awmalik (changes requested):
Thanks again for the PR. Please review my comment regarding the IAM binding for the backup SA.

modules/backup/main.tf
ps-occrp: I think this is better option but I don't understand why it's needed in first place. Workflow is only exporting to the bucket. It does not list or try to get any info about anything else on bucket.

rverenich:
it fails with
```
Access denied for service account ...@gcp-sa-cloud-sql.iam.gserviceaccount.com, permissions missing are: [storage.objects.delete]"
```
permissions missing are: [storage.objects.delete]"

modules/backup/main.tf
rverenich: I don't know why it fails with `permissions missing are: [storage.objects.delete]` , hence objectUser permissions

agree with creating another IAM binding

rverenich:
this workflow started working again without `permissions missing are: [storage.objects.delete]` . It might be some gcloud internal glitch. This PR may be closed for now

PR #613, opened 2024-07-04
feat: adding support for automatic creation of psc consumer

sid-occrp:
When using psc you can optionally create psc consumer for main and read replica in the specified network and subnet.

### DIFF
--- a/examples/postgresql-psc/main.tf
+++ b/examples/postgresql-psc/main.tf
@@ -25,6 +25,22 @@ locals {
   }
 }
 
+module "network-auto-psc" {
+  source  = "terraform-google-modules/network/google"
+  version = "~> 9.0"
+
+  project_id   = var.project_id
+  network_name = "your_network_name"
+
+  subnets = [
+    {
+      subnet_name   = "your-subnet"
+      subnet_ip     = "10.4.0.0/16"
+      subnet_region = "us-central1"
+    }
+  ]
+}
+
 module "pg" {
   source  = "terraform-google-modules/sql-db/google//modules/postgresql"
   version = "~> 20.0"
@@ -85,6 +101,12 @@ module "pg" {
       user_labels       = { bar = "baz" }
     },
   ]
+
+  psc_consumer = {
+    enabled    = true
+    subnet_id  = module.network-auto-psc.subnets_ids[0]
+    network_id = module.network-auto-psc.network_id
+  }
 
   db_name      = var.pg_psc_name
   db_charset   = "UTF8"
--- a/modules/postgresql/main.tf
+++ b/modules/postgresql/main.tf
@@ -43,6 +43,9 @@ locals {
 
   // Force the usage of connector_enforcement
   connector_enforcement = var.connector_enforcement ? "REQUIRED" : "NOT_REQUIRED"
+
+  psc_consumers = var.psc_consumer.enabled ? { for instance in concat([google_sql_database_instance.default], values(google_sql_database_instance.replicas)) : instance.name => instance } : {}
+
 }
 
 resource "random_id" "suffix" {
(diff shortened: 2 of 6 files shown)


modules/postgresql/variables.tf
```diff
@@ -439,3 +439,28 @@ variable "data_cache_enabled" {
...
+    subnet_id               = optional(string, "")
+    network_id              = optional(string, "")
+    enabled                 = optional(bool, false)
+    allow_psc_global_access = optional(bool, false)
+  })
+
+  default = {}
+
+  validation {
+    condition     = (!var.psc_consumer.enabled || (var.psc_consumer.network_id != "" && var.psc_consumer.subnet_id != ""))
```
g-awmalik, line 455: Shouldn't this be based on the error message:
```
var.psc_consumer.enabled && (var.psc_consumer.network_id != "" || var.psc_consumer.subnet_id != "")
```

g-awmalik (commented):
Looks good for the most part. Please take a look at one comment.

modules/postgresql/variables.tf
sid-occrp: Your expression would return true ( valid value ) in case enabled is true and one of network_id or subnet_id is null ( 1 && ( 1 || 0 )). Our expression would return true with the default object and would return true if enabled is true and both network_id or subnet_id are populated. If we "turn off" the module by setting enabled == false the expression would also be true ( expected behaviour ) irrespectively of the input data.

sid-occrp:
@imrannayer @isaurabhuttam  how is my PR looking ?

imrannayer:
@sid-occrp integration test failed

```
Error: "name" ("your_network_name") doesn't match regexp "^(?:[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?)$"

  with module.network-auto-psc.module.vpc.google_compute_network.network,
  on .terraform/modules/network-auto-psc/modules/vpc/main.tf line 21, in resource "google_compute_network" "network":
  21:   name                                      = var.network_name
}
```

modules/postgresql/main.tf
```diff
@@ -320,15 +325,62 @@ resource "google_sql_user" "iam_account" {
   deletion_policy = var.user_deletion_policy
 }
 
+
+resource "google_compute_address" "psc_ilb_consumer_address" {
```
bharathkkb, line 329: For each resource that supports project argument let's explicitly provide it and not depend on inheriting from provider.

examples/postgresql-psc/main.tf
```diff
@@ -25,6 +25,22 @@ locals {
   }
 }
 
+module "network-auto-psc" {
+  source  = "terraform-google-modules/network/google"
+  version = "~> 9.0"
+
+  project_id   = var.project_id
+  network_name = "your_network_name"
```
bharathkkb, line 33: Looks like test failure was due to name constraints ` Returning due to fatal error: FatalError{Underlying: error while running command: exit status 1; 
Error: "name" ("your_network_name") doesn't match regexp "^(?:[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?)$"`

```suggestion
  network_name = "sample-network"
```

modules/postgresql/main.tf
```diff
@@ -320,15 +325,62 @@ resource "google_sql_user" "iam_account" {
...
+  for_each     = local.psc_consumers
+  name         = each.value.dns_name
+  managed_zone = google_dns_managed_zone.psc_dns_zone[each.value.name].name
+  type         = "A"
+  ttl          = 300
+  rrdatas      = [google_compute_address.psc_ilb_consumer_address[each.value.name].address]
+}
+
+
+
```
bharathkkb, line 371: @srinandan suggested we make this optional with default false

PR #615, opened 2024-07-09
feat!: Add enable_google_ml_integration setting for database instance and configurable project roles for default database service account

q2w:
This PR makes below changes,

1. Add `enable_google_ml_integration` setting for database instance. https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/sql_database_instance#enable_google_ml_integration
2. Configurable project roles for default database service account. The default database service account is created for each database instance. To enable database to call ML functions from VertexAI etc, the service account needs suitable project roles. 

This feature has been used in "Gen AI rag with cloud sql" https://github.com/GoogleCloudPlatform/terraform-genai-rag/blob/main/database.tf#L32

### DIFF
--- a/modules/postgresql/main.tf
+++ b/modules/postgresql/main.tf
@@ -73,6 +73,7 @@ resource "google_sql_database_instance" "default" {
     availability_type           = var.availability_type
     deletion_protection_enabled = var.deletion_protection_enabled
     connector_enforcement       = local.connector_enforcement
+    enable_google_ml_integration = var.enable_google_ml_integration
 
     dynamic "backup_configuration" {
       for_each = local.is_secondary_instance ? [] : [var.backup_configuration]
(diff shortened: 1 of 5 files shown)


modules/postgresql/main.tf
```diff
@@ -67,12 +67,13 @@ resource "google_sql_database_instance" "default" {
...
-    availability_type           = var.availability_type
-    deletion_protection_enabled = var.deletion_protection_enabled
-    connector_enforcement       = local.connector_enforcement
+    tier                         = var.tier
+    edition                      = var.edition
+    activation_policy            = var.activation_policy
+    availability_type            = var.availability_type
+    deletion_protection_enabled  = var.deletion_protection_enabled
+    connector_enforcement        = local.connector_enforcement
+    enable_google_ml_integration = var.enable_google_ml_integration
```
bharathkkb, line 76: This was added in [5.25](https://github.com/terraform-google-modules/terraform-google-sql-db/blob/2d5b1154fa6f5450c3bea2d47398f94150196f08/modules/postgresql/versions.tf#L30) so let's bump min versions

https://github.com/terraform-google-modules/terraform-google-sql-db/blob/2d5b1154fa6f5450c3bea2d47398f94150196f08/modules/postgresql/versions.tf#L30

bharathkkb (commented):
LGTM with one request

modules/postgresql/main.tf
q2w: Done.

bharathkkb (approved):

Issue #624, opened 2024-07-30
bug: Invalid Index on disabled default db and additional_databases


Dariusch:
### TL;DR

The latest v21 release introduced a bug regarding a value lookup, `Invalid index`.

### Observed behavior

```
│ Error: Invalid index
│
│   on .terraform/modules/postgresql/modules/postgresql/main.tf line 47, in locals:
│   47:   database_name = var.enable_default_db ? google_sql_database.default[0].name : (length(local.databases) > 0 ? google_sql_database.additional_databases[0].name : "")
│     ├────────────────
│     │ google_sql_database.additional_databases is object with 10 attributes
│
│ The given key does not identify an element in this collection value. An object only supports looking up attributes by name, not by numeric index.
╵
```

### Terraform Configuration

```hcl
module "postgresql" {
  source  = "GoogleCloudPlatform/sql-db/google//modules/postgresql"
  version = "21.0.0"

  ...

  enable_default_db = false

  additional_databases = [{ name = "any_db", charset = "UTF8", collation = "en_US.UTF8" }]

  ...
```

### Terraform Version

```sh
Terraform v1.9.3
on darwin_arm64
+ provider registry.terraform.io/hashicorp/google v5.39.0
+ provider registry.terraform.io/hashicorp/google-beta v5.39.0
...
```

### Additional information

The condition was introduced in https://github.com/terraform-google-modules/terraform-google-sql-db/pull/612/files#diff-5af99345140d405f1b9683608683cd8e45b6cce368ba0446706651ca3a92f041R47

joaoluchetti:
+1

cwillison94:
+1
We are also experiencing this.

imrannayer:
@q2w your PR #612 is causing this error. Can you plz check and send a fix?

imrannayer:
Fixed in #625

PR #625, opened 2024-08-01
bug: fix invalid index with additional database and default database disabled

q2w:
This PR addresses https://github.com/terraform-google-modules/terraform-google-sql-db/issues/624. 

Creation of addition_databases without default database doesn't create an issue of invalid index.

```
module "postgresql" {
  source  = "../terraform-google-sql-db/modules/postgresql"

  project_id = "<PROJECT_ID>"
  name = "database-instance"
  database_version = "POSTGRES_14"
  enable_default_db = false

  additional_databases = [{ name = "any_db_1", charset = "UTF8", collation = "en_US.UTF8" }]
}
```

Note: The PR also disables defaultVerify for postgresql test as it is flaky. No changes in this PR has any impact on the test.

### DIFF
--- a/modules/postgresql/main.tf
+++ b/modules/postgresql/main.tf
@@ -44,7 +44,7 @@ locals {
   // Force the usage of connector_enforcement
   connector_enforcement = var.connector_enforcement ? "REQUIRED" : "NOT_REQUIRED"
 
-  database_name = var.enable_default_db ? google_sql_database.default[0].name : (length(local.databases) > 0 ? google_sql_database.additional_databases[0].name : "")
+  database_name = var.enable_default_db ? var.db_name : (length(var.additional_databases) > 0 ? var.additional_databases[0].name : "")
 }
 
 resource "random_id" "suffix" {
--- a/test/integration/postgresql-backup-provided-service-account/postgresql_backup_provided_service_account_test.go
+++ b/test/integration/postgresql-backup-provided-service-account/postgresql_backup_provided_service_account_test.go
@@ -28,7 +28,7 @@ func TestPostgresqlBackupModuleProvidedServiceAccount(t *testing.T) {
 	mySql := tft.NewTFBlueprintTest(t)
 
 	mySql.DefineVerify(func(assert *assert.Assertions) {
-		mySql.DefaultVerify(assert)
+		// mySql.DefaultVerify(assert)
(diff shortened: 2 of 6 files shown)


modules/postgresql/main.tf
```diff
@@ -44,7 +44,7 @@ locals {
   // Force the usage of connector_enforcement
   connector_enforcement = var.connector_enforcement ? "REQUIRED" : "NOT_REQUIRED"
 
-  database_name = var.enable_default_db ? google_sql_database.default[0].name : (length(local.databases) > 0 ? google_sql_database.additional_databases[0].name : "")
+  database_name = var.enable_default_db ? var.db_name : (length(var.additional_databases) > 0 ? var.additional_databases[0].name : "")
```
imrannayer, line 47: @q2w this will only show the name of first database. Whats the purpose of this database name in output?

```
var.additional_databases[0].name
```

q2w: database_name is part of an output called env_vars which has other database related details. The env_vars can be used by cloud-run, MIG or other modules directly. Eg. https://github.com/q2w/solution-builder-three-tier-web-app/blob/feature/ttwa-adc-default-network-sa-in-cr/infra/main.tf#L39 

```
output "env_vars" {
  description = "Exported environment variables"
  value = {
    "CLOUD_SQL_DATABASE_HOST" : google_sql_database_instance.default.first_ip_address,
    "CLOUD_SQL_DATABASE_CONNECTION_NAME" : google_sql_database_instance.default.connection_name,
  }
}
```

https://github.com/terraform-google-modules/terraform-google-sql-db/blob/master/modules/postgresql/outputs.tf#L152

q2w: @bharathkkb

imrannayer: @q2w thx. But your logic is showing only one database name in case of multiple additional databases. Is that acceptable behavior?

q2w: Yes, We provide one database name (default if present, first additional database otherwise) as part of env_vars. Thus saving users to worry about database name in their application.
Providing all databases in the env_vars doesn't help users as they still need to configure which database to use in the application (which can be done directly).

imrannayer (approved):

PR #629, opened 2024-08-11
chore: Add connections metadata

q2w:
This PR adds connections metadata for postgresql module

### DIFF
--- a/modules/postgresql/metadata.yaml
+++ b/modules/postgresql/metadata.yaml
@@ -212,6 +212,18 @@ spec:
               email = string
             }))
         defaultValue: []
+        connections:
+        - source:
+            source: github.com/terraform-google-modules/terraform-google-service-accounts//modules/simple-sa
+            version: v4.2.3
+          spec:
+            outputExpr: id
+        connections:
+        - source:
+            source: github.com/GoogleCloudPlatform/terraform-google-cloud-run//modules/v2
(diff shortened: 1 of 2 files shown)


modules/postgresql/metadata.yaml
```diff
@@ -212,6 +212,18 @@ spec:
               email = string
             }))
         defaultValue: []
+        connections:
+        - source:
+            source: github.com/terraform-google-modules/terraform-google-service-accounts//modules/simple-sa
+            version: v4.2.3
```
bharathkkb, line 218:
```suggestion
            version: v4.3.0
```

modules/postgresql/metadata.yaml
```diff
@@ -212,6 +212,18 @@ spec:
...
+          spec:
+            outputExpr: id
+        connections:
+        - source:
+            source: github.com/GoogleCloudPlatform/terraform-google-cloud-run//modules/v2
+            version: v0.12.0
```
bharathkkb, line 224:
```suggestion
            version: v0.13.0
```

bharathkkb (approved):
Minor suggestions for updating to releases with the expected outputs

modules/postgresql/metadata.yaml
q2w: Done!

PR #646, opened 2024-09-17
chore: Add alternate defaults for mysql & postgresql blueprint

qz267:
More details in http://docs/document/d/1LMsDRy0zMKQrPUDJ4chlPtqul9bwXs90QH6N8xqZ78I?tab=t.0#heading=h.qx5xjbm73vyx


imrannayer:
@qz267 is there any reason for changing defaults instead of sending values and override it in module call?

qz267:
> @qz267 is there any reason for changing defaults instead of sending values and override it in module call?

Sorry for the confusion, I was trying to surface some of my WIP changes, to ask for early feedbacks, PR title updated.

b/362377949 is the original request, I don't have strong preference about update in variables or override them in the module call, any suggestions are welcomed, thanks!

imrannayer:
@qz267 I think it will be better to take alternate route instead of changing defaults in the module as this change will require all the existing user of the module to adjust code for new defaults.

@bharathkkb can u plz review?

qz267:
I got @q2w suggested using `alt_defaults` instead https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/master/cli/bpmetadata/proto/bpmetadata_ui.proto#L149, WDYT?

modules/mysql/metadata.display.yaml
```diff
@@ -43,6 +43,9 @@ spec:
         backup_configuration:
           name: backup_configuration
           title: Backup Configuration
+          alt_defaults:
+            - type: ALTERNATE_TYPE_UNSPECIFIED
```
bharathkkb, line 47: I think we want these to be `ALTERNATE_TYPE_DC` but will let @q2w confirm

q2w: @qz267 Let's use `ALTERNATE_TYPE_DC ` which we have introduced for ADC!

qz267: DONE

modules/postgresql/metadata.display.yaml
```diff
@@ -40,9 +40,15 @@ spec:
...
           title: Availability Type
+          alt_defaults:
+            - type: ALTERNATE_TYPE_DC
+              value: REGIONAL
         backup_configuration:
           name: backup_configuration
           title: Backup Configuration
+          alt_defaults:
+            - type: ALTERNATE_TYPE_DC
+              value: true
```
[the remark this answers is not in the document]
qz267: Updated in https://github.com/terraform-google-modules/terraform-google-sql-db/pull/646/commits/dba5ec152cf7f3c051eac7b29dce889257d8102b, let me know how it looks for you.

modules/postgresql/metadata.display.yaml
q2w, line 1: `data_cache_enabled ` should also have alt_defaults set to `true`

modules/postgresql/metadata.display.yaml
```diff
@@ -58,6 +68,11 @@ spec:
         database_flags:
           name: database_flags
           title: Database Flags
+          alt_defaults:
+            - type: ALTERNATE_TYPE_DC
+              value:
+                - name: "cloudsql.iam_authentication",
```
q2w, line 74:
```suggestion
                - name: "cloudsql.iam_authentication"
```

modules/postgresql/metadata.display.yaml
```diff
@@ -40,9 +40,19 @@ spec:
...
           title: Availability Type
+          alt_defaults:
+            - type: ALTERNATE_TYPE_DC
+              value: REGIONAL
         backup_configuration:
           name: backup_configuration
           title: Backup Configuration
+          alt_defaults:
+            - type: ALTERNATE_TYPE_DC
+              value:
```
q2w, line 51: The value should be yaml equivalent of `{
  "enabled": true,
  "point_in_time_recovery_enabled": true
}`. I guest it should be like

```
value:
  enabled: true
  point_in_time_recovery_enabled: true
```

modules/mysql/metadata.display.yaml
```diff
@@ -52,9 +60,17 @@ spec:
...
+          alt_defaults:
+            - type: ALTERNATE_TYPE_DC
+              value: true
         database_flags:
           name: database_flags
           title: Database Flags
+          alt_defaults:
+            - type: ALTERNATE_TYPE_DC
+              value:
+                - name: "cloudsql_iam_authentication",
```
q2w, line 72:
```suggestion
                - name: "cloudsql_iam_authentication"
```

q2w (approved):

PR #649, opened 2024-09-23
Update README.md, remove point_in_time_recovery setting as it is not …

eternityduck:
…sufficient for the mysql

Error: point_in_time_recovery_enabled is only available for the following [POSTGRES SQLSERVER]. You may want to consider using binary_log_enabled instead and remove point_in_time_recovery_enabled (removing point_in_time_recovery_enabled and adding binary_log_enabled will enable pitr for MYSQL)


eternityduck:
@isaurabhuttam @imrannayer review pls

eternityduck:
@imrannayer I can see that you have merged branch into mine thus changes probably were applied there

imrannayer:
@eternityduck can we close this one since there is no change needed

PR #654, opened 2024-10-04
chore: Generate output types for sql modules metadata.yaml files

tjy9206:


q2w (approved):
@tjy9206 If one runs `cft blueprint metadata` without generate-output-type flag, would the type info be removed?

bharathkkb (approved):

PR #655, opened 2024-10-10
chore: Update connection metadata

q2w:
Update connection source version to comply with semantic versioning

### DIFF
--- a/modules/postgresql/metadata.yaml
+++ b/modules/postgresql/metadata.yaml
@@ -223,7 +223,7 @@ spec:
         connections:
           - source:
               source: github.com/terraform-google-modules/terraform-google-network//modules/vpc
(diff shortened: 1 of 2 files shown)


ayushmjain (approved):

modules/postgresql/metadata.yaml
```diff
@@ -223,7 +223,7 @@ spec:
         connections:
           - source:
               source: github.com/terraform-google-modules/terraform-google-network//modules/vpc
-              version: v9.1.0
+              version: ">= 9.1.0"
```
bharathkkb, line 226: @q2w Let'ts keep an upper bound on the major version as that may contain breaking changes `~> 9.1`

bharathkkb (approved):

PR #657, opened 2024-10-17
chore: Add Apphub URI output to Cloud SQL modules

tjy9206:


modules/postgresql/outputs.tf
```diff
@@ -152,3 +152,11 @@ output "env_vars" {
     "CLOUD_SQL_DATABASE_NAME" : local.database_name
   }
 }
+
+output "apphub_service_uri" {
+  value = {
+    service_uri = google_sql_database_instance.default.self_link
+    service_id  = google_sql_database_instance.default.name
```
q2w, line 159:
```suggestion
    service_id  = substr("${var.name}-${md5("${var.project_id}")}", 0, 63)
```

modules/mssql/outputs.tf
```diff
@@ -85,3 +85,11 @@ output "primary" {
   description = "The `google_sql_database_instance` resource representing the primary instance"
   sensitive   = true
 }
+
+output "apphub_service_uri" {
+  value = {
+    service_uri = google_sql_database_instance.default.self_link
```
q2w, line 91: Let's also update it as per comment for `postgresql/outputs.tf`

modules/mysql/outputs.tf
```diff
@@ -154,3 +154,11 @@ output "env_vars" {
     "CLOUD_SQL_DATABASE_NAME" : local.database_name
   }
 }
+
+output "apphub_service_uri" {
+  value = {
+    service_uri = google_sql_database_instance.default.self_link
```
q2w, line 160: Let's also update it as per comment for `postgresql/outputs.tf`

modules/mysql/metadata.yaml
```diff
@@ -914,3 +920,12 @@ spec:
       - serviceusage.googleapis.com
       - sqladmin.googleapis.com
       - workflows.googleapis.com
+    providerVersions:
```
q2w, line 923: Thanks for the change! Was this auto-generated?

modules/postgresql/outputs.tf
```diff
@@ -152,3 +152,11 @@ output "env_vars" {
     "CLOUD_SQL_DATABASE_NAME" : local.database_name
   }
 }
+
+output "apphub_service_uri" {
+  value = {
+    service_uri = google_sql_database_instance.default.self_link
```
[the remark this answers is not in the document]
tjy9206: Thanks for this and the service_id comment! Confirmed the uri and id are generated correctly: http://screen/twjYNBTNYY7MDqj

modules/mysql/metadata.yaml
tjy9206: Yes! I think this is generated per @qz267 's PR.

q2w (approved):

modules/mssql/outputs.tf
```diff
@@ -85,3 +85,11 @@ output "primary" {
   description = "The `google_sql_database_instance` resource representing the primary instance"
   sensitive   = true
 }
+
+output "apphub_service_uri" {
+  value = {
+    service_uri = "//cloudsql.googleapis.com/projects/${var.project_id}/instances/${var.name}"
```
bharathkkb, line 91: Rather than interpolating variable names let's extract from computed attribute like self_link

https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/sql_database_instance#self_link

tjy9206: I was using `self_link` initially but @q2w pointed out that the output is not in CAIS format (its `www.googleapis.com/sql/...` instead of `cloudsql.googleapis.com/...`). 

I'm not actually familiar with how these parameters are going to be used in Apphub. Is there a difference between these two representations?

q2w: @tjy9206 I assume Bharat's point is to extract `projects/${var.project_id}/instances/${var.name}` from self_link! Here is an example of self_link `https://sqladmin.googleapis.com/sql/v1beta4/projects/abhiwa-test-30112023/instances/database-instance-2`

tjy9206: Thank you both! Modified accordingly.

PR #661, opened 2024-10-21
feat(TPG>=6.1)!: add support for IAM groups

### LINKED ISSUE #626: Support for iam_groups
### TL;DR

Does this module support IAM groups (https://cloud.google.com/sql/docs/postgres/add-manage-iam-users#add-iam-group-db), or can that be added? It looks like it infers whether an `iam_users` entry is a service account or not, but not whether it is a group or not.

### Terraform Resources

_No response_

### Detailed design

_No response_

### Additional information

My org provides "wrapper" libs around the terraform modules, including this one, so it's not trivial to "add on" a separate call to `resource "google_sql_user"`.

imrannayer:
Fixed #626


imrannayer:
@bharathkkb can u plz review?

modules/mysql/main.tf
```diff
@@ -27,8 +27,9 @@ locals {
   users     = { for u in var.additional_users : u.name => u }
   iam_users = {
     for user in var.iam_users : user.id => {
-      email         = user.email,
+      email         = user.email
       is_account_sa = trimsuffix(user.email, "gserviceaccount.com") == user.email ? false : true
+      type          = user.type != null ? user.type : null
```
q2w, line 32: nit: Should we remove `is_account_sa` altogether since it is utlimately used for finding out type only!
```suggestion
      type          =  trimsuffix(user.email, "gserviceaccount.com") == user.email ? (user.type != null ? user.type : "CLOUD_IAM_USER") : "CLOUD_IAM_SERVICE_ACCOUNT"
```

PR #673, opened 2024-11-28
chore: remove require_ssl from examples

ckluente-1k5:
sorry for this "docs only" PR. But since require_ssl does not do anything, I thought it might be important to fix this before people are surprised that their database still allows unencrypted connections although stated differently in their code.


imrannayer:
@ckluente-1k5 thx for the PR. As of 6.01 require_ssl parameter has been removed. It was removed from the module. We need to remove it from doc and example.

imrannayer:
@ckluente-1k5 integration test is failing. Can you plz replace `ALLOW_UNENCRYPTED_AND_ENCRYPTED` with `ENCRYPTED_ONLY` in [this file](https://github.com/terraform-google-modules/terraform-google-sql-db/blob/main/test/integration/postgresql-with-cross-region-failover/postgresql_cross_region_failover_test.go)

imrannayer (approved):

PR #684, opened 2024-12-20
feat: Remove condition from data.google_compute_zones.available

lukasz-mitka:
Condition causes an issue when `var.zone` is computed in the host Terraform code.

```
  │ Error: Invalid count argument
  │ 
  │   on .terraform/modules/pg/modules/postgresql/read_replica.tf line 26, in data "google_compute_zones" "available":
  │   26:   count   = var.zone == null ? 1 : 0
  │ 
  │ The "count" value depends on resource attributes that cannot be determined
  │ until apply, so Terraform cannot predict how many instances will be
  │ created. To work around this, use the -target argument to first apply only
  │ the resources that the count depends on.
```


imrannayer:
gcbrun

imrannayer (approved):

PR #687, opened 2024-12-25
feat: Add reuse_interval to password validation policy configuration

anluuhoai:


anluuhoai:
friendly ping, Could you review? @imrannayer @q2w

modules/mysql/variables.tf
```diff
@@ -336,6 +336,7 @@ variable "password_validation_policy_config" {
     min_length                  = number
     complexity                  = string
     disallow_username_substring = bool
+    reuse_interval              = number
```
imrannayer, line 339: @anluuhoai if it is not required can u make it optional?

imrannayer: @anluuhoai if this is not required can u plz define as optional

```
reuse_interval = optional(number)
```

anluuhoai: Default is null, if you use it, it needs to be defined, otherwise, it will be null. If it's null, it will not be checked in the box below. reuse_interval should be set the same as min_length, complexity?

Screenshot 2025-02-04 at 09 27 51

anluuhoai: I think reuse_interval is no different from min_length or complexity @imrannayer

anluuhoai: btw, I updated.

imrannayer:
@anluuhoai can you plz lint test?

anluuhoai:
Updated. Please re-run. @imrannayer

imrannayer (approved):

PR #696, opened 2025-01-15
chore: update display metadata and alternate default of deletion_protection for ADC

q2w:
This PR also updates order of variables for `mysql` and `postgresql` modules.

### DIFF
--- a/modules/postgresql/metadata.display.yaml
+++ b/modules/postgresql/metadata.display.yaml
@@ -62,12 +62,14 @@ spec:
         data_cache_enabled:
           name: data_cache_enabled
           title: Data Cache Enabled
+          level: 1
           altDefaults:
             - type: ALTERNATE_TYPE_DC
               value: true
         database_deletion_policy:
           name: database_deletion_policy
           title: Database Deletion Policy
+          level: 1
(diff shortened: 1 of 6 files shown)


modules/postgresql/metadata.display.yaml
```diff
@@ -111,24 +116,29 @@ spec:
...
+          altDefaults:
+            - type: ALTERNATE_TYPE_DC
+              value: false
         deletion_protection_enabled:
           name: deletion_protection_enabled
           title: Deletion Protection Enabled
           altDefaults:
             - type: ALTERNATE_TYPE_DC
-              value: true
+              value: false
```
bharathkkb, line 127: The default for this is already false. IIRC from prev discussions we wanted this to be true in alternate defaults?

q2w: Yes! But this is causing friction when ADC users are modifying the name of sql instance! There application deployment is failing. 
The current ask is to disable it for now and enable at later stage!

Yes, the default is already false! But keeping it here as we would simply make it true when we want it at later stage!

bharathkkb (approved):

PR #704, opened 2025-02-11
feat: add output required for apphub registration for mysql/postgresql module

amrkk123:


modules/postgresql/outputs.tf
```diff
@@ -152,3 +152,11 @@ output "env_vars" {
     "CLOUD_SQL_DATABASE_NAME" : local.database_name
   }
 }
+
+output "apphub_service_uri" {
+  value = {
+    service_uri = "//cloudsql.googleapis.com/projects${element(split("/projects", google_sql_database_instance.default.self_link), 1)}"
+    service_id  = substr(format("%s-%s", var.name, md5(var.project_id)), 0, 63)
```
q2w, line 159: Could you add location field to the output. Similar to https://github.com/GoogleCloudPlatform/terraform-google-cloud-run/blob/main/modules/v2/outputs.tf#L86
It goes for everywhere you are adding apphub_service_uri. Also update the type in the metadata.yaml to reflect new field.

q2w, line 159: Let's also include var.region while creating the service_id as is done elsewhere.

modules/postgresql/outputs.tf
```diff
@@ -152,3 +152,12 @@ output "env_vars" {
...
+    location    = var.location
```
q2w, line 160: I don't think var.location exist in this module. Use var.region for location here. Similar for mysql module. Please validate!

q2w (approved):

PR #717, opened 2025-04-07
feat: import key handle on instance recreate

ps-occrp:
Keyhandle resource cannot be deleted from project so when using autokey if you delete and recreate an instance you should import older keyhandle. This PR implements this functionality in Postgresql module.

Using data block first we check if keyhandle exists or not, if it exists we import it else it will be created.

https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/kms_key_handle

### DIFF
--- a/modules/postgresql/main.tf
+++ b/modules/postgresql/main.tf
@@ -47,6 +47,9 @@ locals {
   database_name = var.enable_default_db ? var.db_name : (length(var.additional_databases) > 0 ? var.additional_databases[0].name : "")
 
   encryption_key = var.encryption_key_name != null ? var.encryption_key_name : var.use_autokey ? google_kms_key_handle.default[0].kms_key : null
+
+  autokey_location = coalesce(var.region, join("-", slice(split("-", var.zone), 0, 2)))
+  autokey_handle   = try({ "handle" = [for handle in data.google_kms_key_handles.existing.key_handles : handle.name if endswith(handle.name, "/${var.name}")][0] }, {})
 }
 
 resource "random_id" "suffix" {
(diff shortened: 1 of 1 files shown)


ps-occrp:
@imrannayer what do you think about this?

modules/postgresql/main.tf
```diff
@@ -214,12 +217,25 @@ resource "google_sql_database_instance" "default" {
   depends_on = [null_resource.module_depends_on]
 }
 
+data "google_kms_key_handles" "existing" {
```
imrannayer, line 220: @ps-occrp instead of searching and importing can we make `google_kms_key_handle.name` an optional variable (kms_key_handle_name) with default value null and also add `create_kms_key_handle` boolean variable which is true by default. 

If use_autokey is true then:

Scenario1:
If user provides a kms_key_handle_name and create_kms_key_handle is false then we will assume it already exists and we will use [google_kms_key_handle](https://registry.terraform.io/providers/hashicorp/google/latest/docs/data-sources/kms_key_handle) data source to find and use it.

Scenario2:
If kms_key_handle_name is provided and create_kms_key_handle is true then we will create key handle using that name 

Scenario 3:
If kms_key_handle_name is not provided and create_kms_key_handle is true then we can just assume we need to create key handle using instance name

imrannayer:
@ps-occrp can u plz review my comments about not using import.

modules/postgresql/main.tf
ps-occrp: @imrannayer In this design fact that key_handle is not really deleted/destroyed is exposed to user (They need to know this and act based on that). On resource level I understand and agree with the decision but on module level I personally would prefer to not expose that to user and make it transparent for user. If you don't agree with it I can adopt change that you suggested. If you agree with the idea but prefer to not use import I think I can workaround that.

imrannayer: I am trying to keep non-standard behavior like import out of the modules. I think it will be a better to allow people to provide handle name. Majority of the customers use random suffix with database due to cloudSql limitation of reusing name within 7 days. Although that limitation is not there but there are lots of folks still deploying cloudsql with random suffix.
Also add this in the documentation and example documentation so users are aware of this limitation. It will be great if you can also add link to GCP/TF provider documentation with this detail.

imrannayer:
@ps-occrp r u still working on this fix? Would be better to get this fix in before next release which will have autikey.

ps-occrp:
@imrannayer sorry for delays, I was not working in mean time. I have updated PR with changes you mentioned.

imrannayer:
@ps-occrp fixed this in #743 . If you delete and recreate instance just set the value of `create_kms_key_handle` to false and it will work fine. If you want you can also pass custom name of key handle in kms_key_handle_name.

PR #718, opened 2025-04-09
Fix: Postgres read-replica "psc_enabled" change never set

philoj:
The `psc_config` dynamic block in resource `google_sql_database_instance.replicas` keeps getting removed by the planner while the defaults are always present. 
This pull request keeps the block with default values when PSC is disabled inside the replica.
No functional changes. Keeps the plan tidy, that's all

Screenshot 2025-04-09 at 11 57 53 AM

### DIFF
--- a/modules/postgresql/read_replica.tf
+++ b/modules/postgresql/read_replica.tf
@@ -64,10 +64,10 @@ resource "google_sql_database_instance" "replicas" {
           }
         }
         dynamic "psc_config" {
-          for_each = ip_configuration.value.psc_enabled ? ["psc_enabled"] : []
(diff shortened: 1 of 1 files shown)


imrannayer:
@philoj this is provider issue. Provider should be able to ignore permadiff for nulls. We can put this workaround in module for the time being.

philoj:
> @philoj this is provider issue. Provider should be able to ignore permadiff for nulls. We can put this workaround in module for the time being.

Didn't know that 🙂. Thank you

imrannayer (approved):

sofuture:
This change seemed to _cause_ a permadiff for those fields, for my configuration at least.

PR #719, opened 2025-04-09
feat: Add output to retrieve passwords by user name for additional_users

nagomiso:
This change allows direct password retrieval by user name.

### DIFF
--- a/modules/postgresql/outputs.tf
+++ b/modules/postgresql/outputs.tf
@@ -115,6 +115,12 @@ output "additional_users" {
   sensitive = true
 }
 
+output "generated_additional_user_passwords" {
+  description = "The auto generated passwords for the additional users"
+  value       = { for u in google_sql_user.additional_users : u.name => u.password }
+  sensitive   = true
+}
+
 output "iam_users" {
   description = "The list of the IAM users with access to the CloudSQL instance"
   value       = var.iam_users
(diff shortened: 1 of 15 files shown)


modules/postgresql/outputs.tf
```diff
@@ -115,6 +115,12 @@ output "additional_users" {
   sensitive = true
 }
 
+output "generated_additional_user_passwords" {
```
imrannayer, line 118: @nagomiso isnt this already covered by this output

```
output "additional_users" {
  description = "List of maps of additional users and passwords"
  value = [for r in google_sql_user.additional_users :
    {
      name     = r.name
      password = r.password
    }
  ]
  sensitive = true
}
```

imrannayer: If you want a Key/Value map in your root module you can convert the output additional_users to a simple Key/Value map by looping through the list:

```
{for x in additional_users : x.name => x.password}
```

nagomiso: Yes. Currently, I'm using the map you suggested by defining it as a local value.

However, I think it would be more user-friendly for module users if I implemented an output in this module that can correctly reference the values ​​created like that map.

imrannayer: @nagomiso lets change `generated_additional_user_passwords` to `additional_user_passwords_map`. Also add in description `Map of auto generated passwords for the additional users`

imrannayer (approved):

PR #728, opened 2025-05-03
feat(TPG>6.17)!: Adding new variables to define failover dr replica

wnugent:


wnugent:
> @wnugent Are you able to complete the steps to sign the CLA? I'm very interested in this feature and would like to see it merged as soon as possible.

@AaronBandt I have signed the CLA already but i just noticed the linting step is failing, i've pushed a fix there that should resolve it

imrannayer:
@wnugent thx for the PR. Can you plz update versions.tf file and set min provider to 6.17 since this feature was released as part of 6.17?

modules/mysql/main.tf
```diff
@@ -64,6 +64,13 @@ resource "google_sql_database_instance" "default" {
   deletion_protection  = var.deletion_protection
   root_password        = var.root_password == "" ? null : var.root_password
 
+  dynamic "replication_cluster" {
```
imrannayer, line 67: @wnugent can u plz add this feature in safer_mysql sub-module?

wnugent: @imrannayer I assume I'd need to set the `master_instance_name` in that module too, as for this to work the target instance would need to be a read replica

wnugent: done

imrannayer (approved):

PR #731, opened 2025-05-19
chore: update current version in readme

TPXP:
Current version stayed on 20.X instead of being updated


imrannayer:
@TPXP can u plz update README to current version is 26.X

Also add a link to it just like we have for others

TPXP:
✅ Updated! I also noticed we were missing some upgrade guides in the list, I've added them

bharathkkb (approved):

Issue #733, opened 2025-05-21
Backups module generates service account with display name that is too long (>100 chars)


OscarVanL:
### TL;DR

The backup service account generated by the `GoogleCloudPlatform/sql-db/google//modules/backup` can have a display_name that is not valid as it is too long. 

```
Error: Error creating service account: googleapi: Error 400: The display_name length (101) is longer than the maximum allowed length 100., badRequest
  with module.cloudsql_backup[0].google_service_account.sql_backup_serviceaccount[0],
  on .terraform/modules/cloudsql_backup/modules/backup/main.tf line 33, in resource "google_service_account" "sql_backup_serviceaccount":
  33: resource "google_service_account" "sql_backup_serviceaccount" {
```

### Expected behavior

The display name should be truncated if too long.

### Observed behavior

My database name is 35 characters long. The `Managed by Terraform - Service account for backup of SQL Instance ` string is 66 characters long. In total this makes 101 characters, which fails the validation check to create the service account.

### Terraform Configuration

```hcl
module "cloudsql_backup" {
  source                 = "GoogleCloudPlatform/sql-db/google//modules/backup"
  version                = "24.0.1"
  export_databases       = [] // exports all databases
  enable_export_backup   = true
  enable_internal_backup = false       // covered by mysql module
  export_schedule        = "0 0 * * *" // at midnight
  export_uri             = google_storage_bucket.db_export_bucket.url
  region                 = "***********"
  project_id             = var.project_id
  sql_instance           = module.db.instance_name
  deletion_protection    = var.backup_deletion_protection
}
```

### Terraform Version

```sh
$ terraform version
Terraform v1.10.1
```

### Terraform Provider Versions

```sh
terraform providers

Providers required by configuration:
.
├── provider[registry.terraform.io/hashicorp/google]
├── module.primary
│   ├── provider[registry.terraform.io/hashicorp/google]
│   ├── provider[registry.terraform.io/hashicorp/random]
│   ├── module.cloudsql_backup
│   │   └── provider[registry.terraform.io/hashicorp/google] >= 6.11.0, < 7.0.0
│   └── module.mysql
│       ├── provider[registry.terraform.io/hashicorp/google-beta] >= 6.1.0, < 7.0.0
│       ├── provider[registry.terraform.io/hashicorp/null] ~> 3.1
│       ├── provider[registry.terraform.io/hashicorp/random] ~> 3.1
[...]

PR #734, opened 2025-05-21
fix: truncate service account display_name in backup module

OscarVanL:
Fixes https://github.com/terraform-google-modules/terraform-google-sql-db/issues/733

My database name is 35 characters long. The `Managed by Terraform - Service account for backup of SQL Instance ` string is 66 characters long. In total this makes 101 characters, which fails the validation check to create the service account:

```
Error: Error creating service account: googleapi: Error 400: The display_name length (101) is longer than the maximum allowed length 100., badRequest
  with module.cloudsql_backup[0].google_service_account.sql_backup_serviceaccount[0],
  on .terraform/modules/cloudsql_backup/modules/backup/main.tf line 33, in resource "google_service_account" "sql_backup_serviceaccount":
  33: resource "google_service_account" "sql_backup_serviceaccount" {
```

This change will truncate the display_name if it exceeds 100 characters.

### DIFF
--- a/modules/backup/main.tf
+++ b/modules/backup/main.tf
@@ -33,7 +33,7 @@ locals {
 resource "google_service_account" "sql_backup_serviceaccount" {
   count        = local.create_service_account ? 1 : 0
   account_id   = trimsuffix(substr("backup-${var.sql_instance}", 0, 28), "-")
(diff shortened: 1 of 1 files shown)


OscarVanL:
@imrannayer What are your thoughts on this, is it ok?

imrannayer (approved):

PR #758, opened 2025-08-07
Output the Service Attachment links for Replicas

yasra002:
# Add service attachment outputs for Postgres replicas

Google's Terraform module for Postgres does not currently expose service attachment outputs for replicas,.This PR adds outputs for the service attachments of Postgres replicas in the Google Cloud SQL Terraform module.
The module already exposes the service attachment for the primary instance, but not for any replicas. This fills that gap so we can reference replica attachments just like the primary.

### DIFF
--- a/modules/postgresql/outputs.tf
+++ b/modules/postgresql/outputs.tf
@@ -88,6 +88,12 @@ output "replicas_instance_server_ca_certs" {
   sensitive   = true
 }
 
+output "replicas_instance_psc_attachments" {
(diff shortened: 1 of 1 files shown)


imrannayer:
@yasra002 thanks for the PR. can u plz fix lint issues. Follow this guid
https://github.com/terraform-google-modules/terraform-google-sql-db/blob/main/CONTRIBUTING.md

yasra002:
@imrannayer  Thanks for the reply. I have now fixed the linting issue.

imrannayer (approved):

PR #765, opened 2025-08-22
chore: fixed escape character issue in regular expressions

amrkk123:


modules/postgresql/metadata.display.yaml
```diff
@@ -235,7 +281,7 @@ spec:
             email:
               name: email
               title: Email
-              regexValidation: ^([a-zA-Z][a-zA-Z0-9_]{0,15}|[a-z0-9._%+-]+@[a-z0-9.-]+\.[a-z]{2,})$
+              regexValidation: "^([a-zA-Z][a-zA-Z0-9_]{0,15}|[a-z0-9._%+-]+@[a-z0-9.-]+\\.[a-z]{2,})$"
```
pawan1210, line 284: Don't think we need to escape anything here. Can you remove the escape backslash.

PR #776, opened 2025-09-25
fix: Add min max validations

amrkk123:


modules/mysql/metadata.display.yaml
```diff
@@ -475,3 +519,17 @@ spec:
           name: zone
           title: Zone
           level: 1
+    runtime:
```
pawan1210, line 522: Is this intended ?

PR #783, opened 2025-10-09
feat: Add psc configs for mssql

msharkey:
Add Private Service Connect (PSC) Configuration Support for MSSQL Module

This pull request adds Private Service Connect (PSC) configuration support to the MSSQL module, bringing it in line with the existing MySQL and PostgreSQL modules that already support PSC.

* This is a non-breaking change - existing configurations will continue to work unchanged
* PSC is disabled by default, requiring explicit opt-in
* Follows the same pattern established by the MySQL and PostgreSQL modules
* This enhancement completes the PSC feature parity across all SQL database modules in this repository.


imrannayer:
@msharkey thanks for the PR. Can you plz fix lint issues? Follow [this doc](https://github.com/terraform-google-modules/terraform-google-sql-db/blob/main/CONTRIBUTING.md)

PR #785, opened 2025-10-30
fix: Enforce backup retention should be > transaction log retention' on mysql replica

diablo02000:
**Description**:

When creating Backup Cloud SQL, the predifined `backup_retention_settings` are the following

```hcl
backup_configuration {
  backup_retention_settings {
    retained_backups = 7
    retention_unit   = "COUNT"
  }
}
```
**Issue**:

If `transaction_log_retention_days` is set to a value greater than the number of `retained_backups`, the following error occurs:

```
Error, failed to update instance settings for : googleapi: Error 400: Invalid request: backup retention should be > transaction log retention., invalid
```

**Proposed Fix**:

Update `retained_backups` value when `transaction_log_retention_days` is greater that predefined value (predefined: 7).


diablo02000:
Hi @imrannayer 

Apologies for the delay. I’ve taken your feedback into account and added a validation block in backup_configuration to ensure that the backup retention period is greater than the transaction log retention period.

I also reviewed the backup_retention_settings configuration for replicas. Since the process value is the same as the master instance, I reused the same code. This approach is simpler than my initial proposal.

PR #807, opened 2026-02-04
fix(postgresql,mysql): ignore psa_write_endpoint to prevent drift when using failover_dr_replica_name

xaviermanachwebedia2:
## Context
When using `failover_dr_replica_name` with PostgreSQL or MySQL modules, 
the Google Cloud SQL API automatically populates the read-only field 
`psa_write_endpoint` in the `replication_cluster` block, causing 
permanent drift.

## Changes
- Add `replication_cluster[0].psa_write_endpoint` to lifecycle.ignore_changes 
  in both postgresql and mysql modules
- This field is read-only and managed by the API (similar to disk_size)


imrannayer:
@xaviermanachwebedia2 thx for the PR. Since these are read only fields it will be more appropriate for provider to not show it as permadiff. I have created an issue for provider team. We can add this workaround for the time being but once provider will fix the issue we can remove it. Here is the link:

https://github.com/hashicorp/terraform-provider-google/issues/25372

PR #809, opened 2026-02-05
feat: add psa_write_endpoint output for postrgres

xaviermanachwebedia2:
## Description

Add `psa_write_endpoint` output to the PostgreSQL module to expose the Private Service Access write endpoint for instances with failover DR replica.

## Motivation

Currently, to retrieve the `psa_write_endpoint` for a Cloud SQL instance with a failover replica, we need to use a data source to re-fetch the instance, which is not optimal:

# Workaround: need to fetch the instance again to get psa_write_endpoint
data "google_sql_database_instance" "default" {
  name    = module.postgresql.instance_name
  project = var.project_id
}

# Access the endpoint
output "psa_endpoint" {
  value = try(data.google_sql_database_instance.default.replication_cluster[0].psa_write_endpoint, null)
}


imrannayer:
@xaviermanachwebedia2 lint test failed.

PR #815, opened 2026-02-26
feat: add database_roles support in mysql/postgres

### LINKED ISSUE #822: Support assignment of roles for DB users in the module
### TL;DR

A new CloudSQL feature launched in January 2026 that allows native + IAM users' database roles to be set in terraform, see [here](https://docs.cloud.google.com/sql/docs/mysql/release-notes#January_26_2026).

Previously, this required you to connect to the database to change the roles, and could not be automated via IaC, which was a maintenance burden.

### Terraform Resources

This can be set via the `google_sql_user`'s `database_roles` field, see [here](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/sql_user#database_roles-1)

### Additional information

This is currently only supported for MySQL and Postgres databases.

eidam:
This PR adds support for [database_roles](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/sql_user#database_roles-1)

fix #822


imrannayer:
@eidam this feature was part of provider version 7.18. Can you plz update versions.tf file for postgres to set min version to 7.18?
Also add details in this doc which will be part of next release

https://github.com/terraform-google-modules/terraform-google-sql-db/blob/main/docs/upgrading_to_sql_db_28.0.md

Can you also add this feature in sub-module safer_mysql?

eidam:
@imrannayer added `safer_mysql`, and modules are already requiring 7.22+

examples/postgresql-public-iam/main.tf
```diff
@@ -77,8 +77,9 @@ module "postgresql-db" {
       email = var.cloudsql_pg_sa
     },
     {
-      id    = "dbadmin",
-      email = "dbadmin@develop.blueprints.joonix.net"
+      id             = "dbadmin",
+      email          = "dbadmin@develop.blueprints.joonix.net"
+      database_roles = ["role1", "role2"]
```
imrannayer: @eidam since these example codes will run as tests can you plz make sure these are real roles otherwise tests will fail?

eidam: @imrannayer done, used `cloudsqlsuperuser` as thats mentioned in both mysql and postgres gcp documentation

eidam:
@imrannayer, what does the `waiting-response` refer to? I believe I answered or followed up on any feedback

pmmmwh:
Bump - it would be really nice to have this added. We have a use case where we want to set accounts with `NOLOGIN` on PG, and only have them as a last resort break-the-glass kind of access that can be switched on. It is impossible to do this with the module now unless we manage the accounts ourselves.

imrannayer (approved):

PR #823, opened 2026-03-13
Support PSC network_attachment_uri setting

dpca:
The `network_attachment_uri` field has been supported as of Google module [v6.48.0](https://github.com/hashicorp/terraform-provider-google/blob/main/CHANGELOG.md#6480-august-12-2025) which is already lower than the versions.tf which requires at least 7.2. This makes the module compatible with PSC outbound networking like in the example here: https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/sql_database_instance#cloud-sql-instance-with-psc-outbound


imrannayer:
@dpca thanks for the PR. Can you plz sign CLA? You should have received an email at @ginkgobioworks.com email address.

PR #825, opened 2026-03-21
Fix incorrect iterator references in connection_pool_config.flags

Rufaidamugalli:
The nested dynamic block was using `flags.name / flags.value`, 
but in Terraform dynamic blocks the current iterated element must be accessed through the iterator’s .value. 
This change updates the block to use `flags.value.name `and `flags.value.value`, fixing the unsupported attribute error in main.tf.


imrannayer:
@Rufaidamugalli can u plz check if #826 fix the problem?

PR #841, opened 2026-05-07
fix: Change 'required' field in connection_pool_config variable to 'default' in mysql and postgresql metadata yaml files

zubitripathy:
Change 'required' field in connection_pool_config variable to 'default' in mysql and postgresql metadata yaml files

Component ingestion and Preview completed successfully with latest changes:
mysql successful preview -> https://paste.googleplex.com/4984958000168960#l=2599
postgresql successful preview-> https://paste.googleplex.com/6608999018528768#l=2676

Successful metadata validation: https://paste.googleplex.com/5932095848448000

### DIFF
--- a/modules/mysql/metadata.yaml
+++ b/modules/mysql/metadata.yaml
@@ -444,7 +444,7 @@ spec:
                 value = string
               })), [])
             })
-        required: true
+        required: false
     outputs:
       - name: additional_users
         description: List of maps of additional users and passwords
--- a/modules/mysql/metadata.yaml
+++ b/modules/mysql/metadata.yaml
@@ -444,7 +444,7 @@ spec:
                 value = string
               })), [])
             })
-        required: false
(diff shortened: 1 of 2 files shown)


modules/mysql/metadata.yaml
```diff
@@ -444,7 +444,7 @@ spec:
                 value = string
               })), [])
             })
-        required: true
```
krprabhat-eng, line 447: don't we need `defaultValue: null` as we have in other optional variable here. I think we don't need `required: false` explicitly if we declare the defaultValue as I can see for other variables.

zubitripathy: Updated the files to use defaultValue: null and removed the redundant required: false to align with the existing pattern.

https://paste.googleplex.com/4984958000168960#l=2599

krprabhat-eng (approved):

PR #843, opened 2026-05-29
feat: Add default value ENCRYPTED_ONLY for ssl_mode

krprabhat-eng:
Add default value ENCRYPTED_ONLY for ADC to fix vulnerabilities.


krprabhat-eng:
Tested standalone deployments of postgresql and mysql with BYOC flow.

ayushmjain (approved):

PR #848, opened 2026-06-25
feat: add data_api_access setting for Cloud SQL Data API

### LINKED ISSUE #845: Adding the `data_api_access` setting to the Postgres CloudSQL Module
### TL;DR

The Postgres CloudSQL module does not expose a property to set the `data_api_access` setting on an instance. It would be convenient to have it added as a property in future module versions.

### Additional information

The `google_sql_database_instance` resource exposes this property. Docs: https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/sql_database_instance#data_api_access-1 .

sanyamjain47:
## Description
  Adds support for the `data_api_access` instance setting, which controls the Cloud SQL Data API (`instances.executeSql`). Requested in #845.

  `google_sql_database_instance` has supported `settings.data_api_access` since provider **7.21.0**; this module pins `>= 7.22`, so no version bump
  is needed.

  ## Changes
  - New `data_api_access` variable (string, default `null`, validated to `ALLOW_DATA_API` / `DISALLOW_DATA_API`) on the `mysql`, `postgresql`,
  `mssql`, and `safer_mysql` modules.
  - Wired into the `settings` block of each primary instance and into `read_replica.tf` for `mysql`/`postgresql` (per-replica override via the
  `read_replicas` object, falling back to the instance value).
  - `safer_mysql` passes the value through to the wrapped `mysql` module.
  - Regenerated `README.md` and updated `metadata.yaml` accordingly.

  Default is `null`, so behaviour is unchanged unless the value is set.

  ## Impact
  - No breaking changes; the new variable is optional and defaults to `null` (API default `DISALLOW_DATA_API`).
  - No performance/resource impact.

  Closes #845


sanyamjain47:
still relevant, please review
@ayushmjain @imrannayer @krprabhat-eng

